Issue #5, opened 2017-04-11
Make it possible to store some state together with the TaskInstance


tedsteen:
To be able to have some state associated with a task (without the need to keep a separate persistent store with task id -> state) it would be nice if one could store some state with a task instance.

#### Use case
I want to schedule an event to be published in the future.
Today I'd have to schedule a task and store the event together with the task id.
If I could store some state with my task I could put my event in that place and pick it up when the task executes.
```
scheduler.scheduleForExecution(triggerDateTime, new EventTask(taskName).instance(instanceId, "my state"));
....
@Override
public void execute(TaskInstance taskInstance, ExecutionContext executionContext) {
  //Do my thing with taskInstance.getState();
}
```

It could be a simple string and the user would have to handle serialization, or it could be more involved with serializers etc as part of the library.

tedsteen:
An alternative, not to break the current api:
`new EventTask(taskName).instance(instanceId).withState("my state")`

Additions to TaskInstance:
```
private final String state;

private TaskInstance(Task task, String id, String state) {
	this.task = task;
	this.id = id;
	this.state = state;
}

public TaskInstance withState(String serializedState) {
	return new TaskInstance(task, id, serializedState);
}
```

kagkarlsson:
Another good suggestion :). This feature has been pretty high on my todo-list as well. Primarily for the type of use-case you describe, to make one-time tasks more useful. The simplest thing would be to add a column to the table, an optional `state` field. Since it is optional, maybe we can add another method `instance(instanceId, state)`. 

I am unsure about what type to use for the state field.. need to think a bit about that. Would `byte[]` be slightly better than `String` perhaps, to encoding issues?
It's a goal to keep it simple, so maybe moving unmarshalling/deserialization to the "outside" would be a good thing. Just making it a bit harder to use.. 

Side-note: For extremely simple state, say a `long` primary key, it is possible to store it as the `instanceId´.

tedsteen:
>I am unsure about what type to use for the state field.. need to think a bit about that. 

Either that, or use string and enforce a serializer that serializes to string to keep it simple.
There are also other [issues](http://stackoverflow.com/questions/3164072/large-objects-may-not-be-used-in-auto-commit-mode) when not using transactions in combination with binary large objects in postgre.

My suggested PR uses String. I'd say this is a simpler solution, but quite opinionated. All ears for suggestions tho :)

>Would byte[] be slightly better than String perhaps, to encoding issues?

[...]

PR #8, opened 2017-04-12
Store state as byte[] together with task instance. #5

tedsteen:
Alternate suggestion for #5 using byte[] as state instead of String

### DIFF
--- a/src/main/java/com/github/kagkarlsson/scheduler/JdbcTaskRepository.java
+++ b/src/main/java/com/github/kagkarlsson/scheduler/JdbcTaskRepository.java
@@ -16,7 +16,6 @@
 package com.github.kagkarlsson.scheduler;
 
 import com.github.kagkarlsson.jdbc.JdbcRunner;
-import com.github.kagkarlsson.jdbc.Mappers;
 import com.github.kagkarlsson.jdbc.ResultSetMapper;
 import com.github.kagkarlsson.jdbc.SQLRuntimeException;
 import com.github.kagkarlsson.scheduler.task.Execution;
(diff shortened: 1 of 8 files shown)


kagkarlsson:
A few comments, up for discussion if you disagree :)

* I feel maybe `data` is a better term than `state`, `state` almost implies that it will change over time (for field-name and column-name)
* The `withState(...)` is perhaps unnecessary if there is a constructor taking a state-argument
* `byte[]` feels slightly more honest/clear, InputStream might leave some questions like "is it lazy?", and String is not a natural type to keep "data" in. However, it might be a good idea to look into creating a new type of `Task` that does serialization and is "type-safe".

tedsteen:
>I feel maybe data is a better term than state, state almost implies that it will change over time (for field-name and column-name)

Agreed!

>The withState(...) is perhaps unnecessary if there is a constructor taking a state-argument

This crossed my mind as well. Will amend.

>byte[] feels slightly more honest/clear, InputStream might leave some questions like "is it lazy?", and String is not a natural type to keep "data" in.

Agree here as well, it's def more honest with byte[].

>However, it might be a good idea to look into creating a new type of Task that does serialization and is "type-safe".

This sounds like an interesting idea, will give it a shot!

tedsteen:
Did amendments for the points you brought up. The last point I did in a separate PR

kagkarlsson:
Looks good. I think this PR can be merged :)

tedsteen:
Cool, this one or the one with the type information on the task instance? (#9)

kagkarlsson:
See comment in #9 , I think we can do the type-safe task in a separate PR

Issue #4, opened 2017-04-11
Provide a way to cancel and reschedule tasks


tedsteen:
Make it possible to cancel and reschedule given a task name and an instance Id.
So basically something like this:
```
final String taskName = "myTaskName";
final OneTimeTask task = new MyOneTimeTask(taskName);
final String instanceId = "myInstanceId";

scheduler.scheduleForExecution(Instant.now().plusSeconds(5), task.instance(instanceId));

scheduler.reSchedule(taskName, instanceId, Instant.now().plusSeconds(15));
//Or
scheduler.cancelSchedule(taskName, instanceId);
```

kagkarlsson:
Excellent suggestion :). Will try to add support for this in the coming week.

tedsteen:
>Excellent suggestion :).

Thanks!
>Will try to add support for this in the coming week.

If you could give me a few hints how to go about this (simple list of basic steps needed to achieve this) I could make a PR for you to approve or continue working on.

kagkarlsson:
I think we need to 
* Add the methods to `SchedulerClient` interface, and implement in `Scheduler`
* Easiest thing is to use existing methods for reschedule and remove in JdbcTaskRepository, after using getExecution to find any execution matching task+instance

kagkarlsson:
Merged PR to master. Will hopefully get to do a new release soon, just have to update some docs first.

kagkarlsson:
This feature was implemented in version 2.0 thanks to the PR from @tedsteen .
Closing the issue.

PR #10, opened 2017-04-18
Reschedule and cancel

tedsteen:
Suggestion for #4

### DIFF
--- a/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
+++ b/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
@@ -32,11 +32,13 @@ import java.util.stream.Collectors;
 
 import static com.github.kagkarlsson.scheduler.ExecutorUtils.defaultThreadFactoryWithPrefix;
 
-public class Scheduler extends SchedulerClient.StandardSchedulerClient {
+public class Scheduler implements SchedulerClient {
 
 	private static final Logger LOG = LoggerFactory.getLogger(Scheduler.class);
 	public static final Duration SHUTDOWN_WAIT = Duration.ofMinutes(30);
+	private final SchedulerClient delegate;
 	private final Clock clock;
+	private final TaskRepository taskRepository;
 	private final ExecutorService executorService;
 	private final Waiter waiter;
 	private final List<OnStartup> onStartup;
@@ -62,8 +64,8 @@ public class Scheduler extends SchedulerClient.StandardSchedulerClient {
 
 	Scheduler(Clock clock, TaskRepository taskRepository, int maxThreads, ExecutorService executorService, SchedulerName schedulerName,
 			  Waiter waiter, Duration heartbeatInterval, StatsRegistry statsRegistry, List<OnStartup> onStartup) {
-		super(taskRepository);
 		this.clock = clock;
+		this.taskRepository = taskRepository;
 		this.executorService = executorService;
 		this.waiter = waiter;
 		this.onStartup = onStartup;
@@ -75,6 +77,7 @@ public class Scheduler extends SchedulerClient.StandardSchedulerClient {
 		this.detectDeadExecutor = Executors.newSingleThreadExecutor(defaultThreadFactoryWithPrefix(schedulerName.getName() + "-detect-dead-"));
 		this.updateHeartbeatExecutor = Executors.newSingleThreadExecutor(defaultThreadFactoryWithPrefix(schedulerName.getName() + "-update-heartbeat-"));
 		executorsSemaphore = new Semaphore(maxThreads);
+		delegate = new StandardSchedulerClient(taskRepository);
 	}
 
 	public void start() {
@@ -119,6 +122,26 @@ public class Scheduler extends SchedulerClient.StandardSchedulerClient {
 			LOG.warn("Scheduler stopped, but some tasks did not complete. Was currently running the following executions:\n{}",
 					new ArrayList<>(currentlyProcessing.keySet()).stream().map(Execution::toString).collect(Collectors.joining("\n")));
 		}
+	}
+
+	@Override
+	public void scheduleForExecution(Instant executionTime, TaskInstance taskInstance) {
+		this.schedule(executionTime, taskInstance);
+	}
+
+	@Override
+	public void schedule(Instant executionTime, TaskInstance taskInstance) {
+		this.delegate.schedule(executionTime, taskInstance);
+	}
+
+	@Override
+	public void reschedule(String taskName, String instanceId, Instant newExecutionTime) {
+		this.delegate.reschedule(taskName, instanceId, newExecutionTime);
+	}
+
+	@Override
+	public void cancel(String taskName, String instanceId) {
+		this.delegate.cancel(taskName, instanceId);
 	}
 
 	public List<CurrentlyExecuting> getCurrentlyExecuting() {
--- a/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
+++ b/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
@@ -32,12 +32,11 @@ import java.util.stream.Collectors;
 
 import static com.github.kagkarlsson.scheduler.ExecutorUtils.defaultThreadFactoryWithPrefix;
 
-public class Scheduler implements SchedulerClient {
+public class Scheduler extends SchedulerClient.StandardSchedulerClient {
 
 	private static final Logger LOG = LoggerFactory.getLogger(Scheduler.class);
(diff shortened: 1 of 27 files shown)


src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
```diff
@@ -32,12 +32,11 @@
 
 import static com.github.kagkarlsson.scheduler.ExecutorUtils.defaultThreadFactoryWithPrefix;
 
-public class Scheduler implements SchedulerClient {
+public class Scheduler extends SchedulerClient.StandardSchedulerClient {
```
kagkarlsson: What is the reason SchedulerClient interface was changed for extends here? The idea behind SchedulerClient is that you are able to "downgrade" the full Scheduler object to only a SchedulerClient, not exposing everything

kagkarlsson:
Made a small comment. Will have a more thorough look later

kagkarlsson:
Except for my comment, looks good on first glance

src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
tedsteen: Because the implementation for was the same for both of them (`Scheduler` and `StandardSchedulerClient`).

Given the fact that `StandardSchedulerClient` implements `SchedulerClient` you will still be able to "downgrade" the full Scheduler object to only a SchedulerClient.

My reasoning was that if `Scheduler` extends `StandardSchedulerClient`:
* they will share the implementation so we don't repeat ourselves
* No new functionality will be exposed on `Scheduler` as `StandardSchedulerClient` only implements `SchedulerClient`

WDYT?

kagkarlsson: Ah, you are correct, and I agree with your reasoning. For some reason, that extends still makes me itch :). Another option is implementing the interface, but delegating to an internal `StandardSchedulerClient`. I know it is pretty cosmetic though..

kagkarlsson:
I see now that there is a potential issue with this approach for cancel/reschedule. If the execution is executed by the scheduler after the `getExecution` but before the `cancel/reschedule`, the call will fail due to the version number being increased (the optimistic locking db-scheduler uses heavily). Also, modifying it while it is "locked" for execution could lead to some unpredictable result, so we should probably only allow the cancel/reschedule if it is not currently executing.

Not sure if there is a way around having custom reschedule and cancel methods.. enforcing `picked = false`. Maybe if we check for `picked=false` after the `getExecution`, and just let it fail if it has been picked for execution in the mean time. The only thing is that currently, the exception message is "Expected one execution to be updated, but updated X. Indicates a bug.", which is kinda misleading for this case that might actually happen. 

Sorry, this was kinda train-of-thoughty :). I will try and make comments for the code lines.

src/main/java/com/github/kagkarlsson/scheduler/SchedulerClient.java
```diff
@@ -27,9 +28,13 @@
 public interface SchedulerClient {
 	
 	void scheduleForExecution(Instant exeecutionTime, TaskInstance taskInstance);
```
kagkarlsson: Maybe we should do some renaming here. 
How about:
* schedule
* reschedule
* cancel

And maybe a default method for the old scheduleForExecution, possibly deprecated?

src/main/java/com/github/kagkarlsson/scheduler/SchedulerClient.java
```diff
@@ -59,7 +64,27 @@ public SchedulerClient build() {
...
+			} else {
+				throw new RuntimeException(String.format("Could not reschedule - no task with name '%' and id '%s' was found." , taskName, instanceId));
+			}
+		}
+
+		@Override
+		public void cancelSchedule(String taskName, String instanceId) {
+			Optional<Execution> execution = taskRepository.getExecution(taskName, instanceId);
+			if(execution.isPresent()) {
+				taskRepository.remove(execution.get());
```
kagkarlsson: if `picked = true` here, throw exception about not allowed to cancel executions that are currently executing

src/main/java/com/github/kagkarlsson/scheduler/SchedulerClient.java
```diff
@@ -59,7 +64,27 @@ public SchedulerClient build() {
...
 				TaskInstance taskInstance) {
 			taskRepository.createIfNotExists(new Execution(exeecutionTime, taskInstance));
-		} 
+		}
+
+		@Override
+		public void reschedule(String taskName, String instanceId, Instant newExecutionTime) {
+			Optional<Execution> execution = taskRepository.getExecution(taskName, instanceId);
+			if(execution.isPresent()) {
+				taskRepository.reschedule(execution.get(), newExecutionTime, null, null);
```
kagkarlsson: See comment for cancel

src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
tedsteen: I see. Which one would you prefer?
* delegating to internal `StandardSchedulerClient`
* make an `AbstractSchedulerClient` and extend from that in both `StandardSchedulerClient` and `Scheduler`

src/main/java/com/github/kagkarlsson/scheduler/SchedulerClient.java
tedsteen: Sounds like a plan!

src/main/java/com/github/kagkarlsson/scheduler/SchedulerClient.java
tedsteen: 👍

src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
kagkarlsson: I think I would prefer keeping the interface and delegate to internal `StandardSchedulerClient`

tedsteen: I think I'd prefer that as well actually. I'll get crackin' :)

src/main/java/com/github/kagkarlsson/scheduler/SchedulerClient.java
```diff
@@ -25,11 +26,23 @@
...
+	 * This method has been renamed and will be removed in a future version.
+	 *
+	 * @deprecated use {@link #schedule(Instant, TaskInstance)} instead.
+	 */
+	@Deprecated
+	void scheduleForExecution(Instant executionTime, TaskInstance taskInstance);
+
+	void schedule(Instant executionTime, TaskInstance taskInstance);
+
+	void reschedule(String taskName, String instanceId, Instant newExecutionTime);
```
kagkarlsson: Changes looking good. Starting to feel I lack a type here, a `TaskInstanceId` or something. Previously that was just a `TaskInstance`, but now a `TaskInstance` is something more, it can have data as well. I will think about it, maybe I will change it in a later version :)

tedsteen: You're right. It seems to be quite intrusive so, as you say, might be better to do in another PR/version.

tedsteen: Btw, I did a small PR in my fork to try it out. If you like it I can merge and it'll be part of this PR.
https://github.com/tedsteen/db-scheduler/pull/1

kagkarlsson: Nice! Doesn't look like it required too many changes to introduce the type. Might as well make it part of this PR.

Was thinking we should also change the order of parameters of the `schedule` method, for symmetry...  
```
void schedule(TaskInstance taskInstance, Instant executionTime);
```

kagkarlsson:
Looks good, thanks for the PR! :)

Merging to master

tedsteen:
Woho!

When do you think you could make a release so it could be picked up through maven?

kagkarlsson:
Just deployed a 1.10-SNAPSHOT, but I'm assuming you mean release-version. I
think I just need to update a couple of things in the readme, then it
should be good to go. Hopefully within a couple of days.. depends on how
much time I can find :)

2017-05-02 21:30 GMT+02:00 Ted Steen <notifications@github.com>:

tedsteen:
Actually, 1.10-SNAPSHOT will give me the master right?
In that case, there's no urgency at all from my end. I'll just use the snapshot.

kagkarlsson:
Yeah, it's from master. Let me know if it is not available via maven


tir. 2. mai 2017 kl. 21.39 skrev Ted Steen <notifications@github.com>:

PR #13, opened 2017-04-25
Tasks where made type safe.

tedsteen:
See https://github.com/kagkarlsson/db-scheduler/issues/5#issuecomment-296817502 #5

### DIFF
--- a/src/main/java/com/github/kagkarlsson/scheduler/task/Task.java
+++ b/src/main/java/com/github/kagkarlsson/scheduler/task/Task.java
@@ -67,7 +67,17 @@ public abstract class Task<T> implements ExecutionHandler<T> {
 	public interface Serializer<T> {
 		byte[] serialize(T data);
 		T deserialize(byte[] serializedData);
+        Serializer NO_SERIALIZER = new Serializer<Void>() {
+            @Override
+            public byte[] serialize(Void data) {
+                return new byte[0];
+            }
 
+            @Override
+            public Void deserialize(byte[] serializedData) {
+                return null;
+            }
+        };
 		Serializer JAVA_SERIALIZER = new Serializer<Object>() {
 			public byte[] serialize(Object data) {
 				if(data == null) return null;
--- a/src/main/java/com/github/kagkarlsson/scheduler/task/Task.java
+++ b/src/main/java/com/github/kagkarlsson/scheduler/task/Task.java
@@ -15,30 +15,40 @@
  */
 package com.github.kagkarlsson.scheduler.task;
 
-public abstract class Task implements ExecutionHandler {
+import java.io.*;
+
+import static com.github.kagkarlsson.scheduler.task.Task.Serializer.JAVA_SERIALIZER;
+
+public abstract class Task<T> implements ExecutionHandler<T> {
 	protected final String name;
 	private final CompletionHandler completionHandler;
 	private final DeadExecutionHandler deadExecutionHandler;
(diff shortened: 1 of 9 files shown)


tedsteen:
Now, I know this is not exactly what you had in mind [here](https://github.com/kagkarlsson/db-scheduler/issues/5#issuecomment-296817502), but I'd like to hear your opinion on this one.

It's a more general solution and it doesn't add as many new methods on Task. Only one (getSerializedData())

kagkarlsson:
Yes I agree with you, I think this is a better solution than what I suggested. This is the proper generics-solution :)

However, for the `RecurringTask` I think the default is that you don't use the data concept. At least that is the case for where I have used them so far. So maybe some sort of `Null` type should be introduced to be used in these cases. Or possibly `RecurringTask` could be made none-generic by default (`RecurringTask extends Task<Null>`)...

src/main/java/com/github/kagkarlsson/scheduler/task/Task.java
```diff
@@ -53,5 +63,31 @@ public String toString() {
 		return "Task " +
 				"task=" + getName();
 	}
+
+	public interface Serializer<T> {
+		byte[] serialize(T data);
+		T deserialize(byte[] serializedData);
+
+		Serializer JAVA_SERIALIZER = new Serializer<Object>() {
+			public byte[] serialize(Object data) {
```
kagkarlsson: Should this have been made generic as well? Trying to figure out where the cast is otherwise.. :)

tedsteen:
>However, for the RecurringTask I think the default is that you don't use the data concept. At least that is the case for where I have used them so far. So maybe some sort of Null type should be introduced to be used in these cases. Or possibly RecurringTask could be made none-generic by default (RecurringTask extends Task<Null>)...

Makes sense 👍

src/main/java/com/github/kagkarlsson/scheduler/task/Task.java
tedsteen: Good catch! Will amend!

tedsteen: Ah, can't do that here, it's a static field on the interface.. Would have to create a new instance per type, but that feels a bit too much perhaps?

kagkarlsson:
Btw, completely unrelated to this PR, thought you had a curious name for living on Malta, so I did a check on LinkedIn. If it was correct, it seems we have some common connections (Hong o HP).. I also went to LiTH, finished in 2007 :)

kagkarlsson:
One instance per type should be fine I think

tir. 25. apr. 2017 kl. 21.48 skrev Ted Steen <notifications@github.com>:

tedsteen:
>Btw, completely unrelated to this PR, thought you had a curious name for living on Malta, so I did a check on LinkedIn. If it was correct, it seems we have some common connections (Hong o HP).. I also went to LiTH, finished in 2007 :)

Haha, look at that! Yeah, that's me :) Both HP and Hong where good friends of mine! I connected on linked in!

src/main/java/com/github/kagkarlsson/scheduler/task/Task.java
kagkarlsson: New instance per type is fine

kagkarlsson:
Will look at merging this soon I hope. Do you have any outstanding things you want to do?

Issue #24, opened 2018-04-12
[Feature] A way to get all executions via the Scheduler


MarkvV:
It would be nice to get all scheduled executions via the `Scheduler` from the `TaskRepository`. Currently there is already a `Scheduler#getCurrentlyExecuting` but this returns the **_currently_** executing tasks. I propose  the `TaskRepository` to have a `public List<Execution> getExecutions()`. This could be helpful when a bunch of executions are scheduled and you want to cancel them at a later time with the use of `scheduler.cancel(taskInstanceId)` without keeping the `taskInstanceId` in memory / in a different storage. The thing is that the `taskInstanceId` is a random generated id and the handle I have to the task is in the data of the task itself. Not something I can query or cancel the task on.

Tell me if you have any questions / remarks. Thank you!

kagkarlsson:
Sounds like a neat feature, as long as the total number of executions is not too big. Pull requests are welcome :)

MarkvV:
[PR](https://github.com/kagkarlsson/db-scheduler/pull/25) is open :-).

kagkarlsson:
Merged to master.

kagkarlsson:
Released in 3.1, closing.

kagkarlsson:
🎉 This issue has been resolved in `v16.9.0` ([Release Notes](https://github.com/kagkarlsson/db-scheduler/releases/tag/v16.9.0))

PR #25, opened 2018-04-13
Issue 24 get all executions

MarkvV:
In general I did an implementation of issue #24, including unit tests. 

Additionally I had to update some of the dependencies in order to get the unit tests running on my machine (Windows). This also lead to updating the notice file as the external libraries used new libraries at their part.

The implementation is fairly simple, I only wonder if I should filter out due tasks when using `Repository#getScheduled`. So that is open for discussion, but I think the current implementation suits my requirement(s).

### DIFF
--- a/src/main/java/com/github/kagkarlsson/scheduler/JdbcTaskRepository.java
+++ b/src/main/java/com/github/kagkarlsson/scheduler/JdbcTaskRepository.java
@@ -40,7 +40,7 @@ import static java.util.Optional.ofNullable;
 public class JdbcTaskRepository implements TaskRepository {
 
 	private static final Logger LOG = LoggerFactory.getLogger(JdbcTaskRepository.class);
-	private static final int MAX_DUE_RESULTS = 10_000;
+	private static final int MAX_RESULTS = 10_000;
 	private final TaskResolver taskResolver;
 	private final SchedulerName schedulerSchedulerName;
 	private final JdbcRunner jdbcRunner;
@@ -92,7 +92,12 @@ public class JdbcTaskRepository implements TaskRepository {
 
 	@Override
 	public List<Execution> getDue(Instant now) {
-		return getDue(now, MAX_DUE_RESULTS);
+		return getDue(now, MAX_RESULTS);
+	}
+
+	@Override
+	public List<Execution> getScheduled() {
+		return getScheduled(MAX_RESULTS);
 	}
 
 	public List<Execution> getDue(Instant now, int limit) {
@@ -101,6 +106,17 @@ public class JdbcTaskRepository implements TaskRepository {
 				(PreparedStatement p) -> {
 					p.setBoolean(1, false);
 					p.setTimestamp(2, Timestamp.from(now));
+					p.setMaxRows(limit);
+				},
+				new ExecutionResultSetMapper()
+		);
+	}
+
+	public List<Execution> getScheduled(int limit) {
+		return jdbcRunner.query(
+				"select * from scheduled_tasks where picked = ? order by execution_time asc",
+				(PreparedStatement p) -> {
+					p.setBoolean(1, false);
 					p.setMaxRows(limit);
 				},
 				new ExecutionResultSetMapper()
--- a/src/main/java/com/github/kagkarlsson/scheduler/JdbcTaskRepository.java
+++ b/src/main/java/com/github/kagkarlsson/scheduler/JdbcTaskRepository.java
@@ -100,6 +100,11 @@ public class JdbcTaskRepository implements TaskRepository {
 		return getScheduled(MAX_RESULTS);
 	}
 
+	@Override
+	public List<Execution> getScheduledByTaskName(String taskName) {
+		return getScheduledByTaskName(taskName, MAX_RESULTS);
+	}
+
 	public List<Execution> getDue(Instant now, int limit) {
 		return jdbcRunner.query(
 				"select * from scheduled_tasks where picked = ? and execution_time <= ? order by execution_time asc",
@@ -117,6 +122,18 @@ public class JdbcTaskRepository implements TaskRepository {
 				"select * from scheduled_tasks where picked = ? order by execution_time asc",
 				(PreparedStatement p) -> {
 					p.setBoolean(1, false);
+					p.setMaxRows(limit);
+				},
+				new ExecutionResultSetMapper()
+		);
+	}
+
+	public List<Execution> getScheduledByTaskName(String taskName, int limit){
+		return jdbcRunner.query(
+				"select * from scheduled_tasks where picked = ? and task_name = ? order by execution_time asc",
+				(PreparedStatement p) -> {
+					p.setBoolean(1, false);
+					p.setString(2, taskName);
 					p.setMaxRows(limit);
 				},
 				new ExecutionResultSetMapper()
--- a/src/main/java/com/github/kagkarlsson/scheduler/JdbcTaskRepository.java
+++ b/src/main/java/com/github/kagkarlsson/scheduler/JdbcTaskRepository.java
@@ -25,12 +25,16 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.sql.DataSource;
-import java.sql.*;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Timestamp;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.function.Consumer;
 import java.util.function.Supplier;
 
 import static com.github.kagkarlsson.scheduler.StringUtils.truncate;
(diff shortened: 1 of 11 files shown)


kagkarlsson:
Excellent, will try and have a look at it tonight. Looks good on first glance :)

src/main/java/com/github/kagkarlsson/scheduler/ScheduledExecution.java
```diff
@@ -0,0 +1,34 @@
...
+    private final Execution execution;
+
+    public ScheduledExecution(Execution execution) {
+        this.execution = execution;
+    }
+
+    public TaskInstance getTaskInstance() {
+        return execution.taskInstance;
+    }
+
```
kagkarlsson: Didn't you say your handle was in the data of the execution? Wouldn't you want to expose that as well for your requirement?  (not sure how I would do that though, since task is generic. you would have to do a check on task-name, and have a method like `<T> T getData(Class<T> dataClass)` to get the deserialized data)

For your requirement, maybe it would make more sense to get all the executions for a specific task only, then the data-class is the same for all? 

Up for discussion..

src/main/java/com/github/kagkarlsson/scheduler/JdbcTaskRepository.java
```diff
@@ -107,6 +112,17 @@ public boolean createIfNotExists(Execution execution) {
 		);
 	}
 
+	public List<Execution> getScheduled(int limit) {
```
kagkarlsson: I have been thinking... :)

Considering that we are including the data (I assume?), maybe it is better to have a streaming interface, something like
`getScheduled(Instant now, Consumer<Execution> consumer)`

The method would then call the consumer for every row in the result-set, so that it would not have to read everything into memory first. That would be a "client" decision, he could if he wanted to and new that is was safe memory-wise.

MarkvV:
@kagkarlsson thanks for your comments, will come back to you at monday :)

src/main/java/com/github/kagkarlsson/scheduler/ScheduledExecution.java
MarkvV: Good point, because I was actually planning to filter on task name and type cast it on the client side. So I do agree that it seems desirable to get all the executions for a specific task. So maybe something like the following?
```java
public class ScheduledExecution<DATA_TYPE> {

    private final Execution execution;

    public ScheduledExecution(Execution execution) {
        this.execution = execution;
    }

    public TaskInstance getTaskInstance() {
        return execution.taskInstance;
    }

    public DATA_TYPE getData() {
        return (DATA_TYPE) this.execution.taskInstance.getData();
    }
}
```
Where the `Scheduler` could have both:
```java
public <T> List<ScheduledExecution<T>> getScheduled() {}
public <T> List<ScheduledExecution<T>> getScheduled(String taskName) {}
```

The client could then do:
```java
// Get all without type identification 
// Here the client could still type-cast if needed
// i.e. String data = (String) scheduler.getScheduled().get(0).getData();
List<ScheduledExecution<Object>> executions1 = scheduler.getScheduled();

// Seems less unlikely, but it could be that all executions are of 1 specific data type.
List<ScheduledExecution<String>> executions2 = scheduler.getScheduled();

// Get by specific task name.
List<ScheduledExecution<String>> executions3 = scheduler.getScheduled(TASK);
```

Let me know what you think and I will get it implemented :-)

src/main/java/com/github/kagkarlsson/scheduler/JdbcTaskRepository.java
MarkvV: Fair enough, seems like a good use for `Consumer<>`, will take this along :).

src/main/java/com/github/kagkarlsson/scheduler/ScheduledExecution.java
kagkarlsson: With the consumer-suggestion this becomes something like this I suppose?

```
List<ScheduledExecution<SomeId>> all = new ArrayList<>();
scheduler.<SomeId>getScheduled(TASK_NAME, add:add)

all.stream()
        .filter(data -> data.getData().getSomeId().equals(some_other_id))
        .map(ScheduledExecution::getTaskInstance)
        .forEach(scheduler::cancel);```

Comment (kagkarlsson) [reply src/main/java/com/github/kagkarlsson/scheduler/ScheduledExecution.java]: And also not sure if maybe we should send in the dataclass as a parameter. Would be easier to document..

`void getScheduled(Consumer<ScheduledExecution<Object>> executionConsumer)`
`<T> void getScheduledForTask(String taskName, T dataClass, Consumer<ScheduledExecution<T>> executionConsumer)` 

This would be a bit prettier if I had some sort of static `TaskDescriptor` class, holding name and dataclass-type. A refactoring I am considering, but not prio yet..  (out of scope here)

Comment (MarkvV) [reply src/main/java/com/github/kagkarlsson/scheduler/ScheduledExecution.java]: Yeah you are right, I didn't use the consumer in this example.

You could have two `getScheduled()`-methods one with the consumer and thus no return type and one without the consumer with a `List<>` as return type.. However, one could also implement the consumer like (without the need of adding it to a list all together):
```java
scheduler.<SomeId>getScheduled(ActivateBeaconTask.CREATE_BEACON_TASK, scheduledExecution -> {
    if (scheduledExecution.getData().getSomeId().equals(some_other_id)) {
        scheduler.cancel(scheduledExecution.getTaskInstance());
    }
});
```
so I leave that one up to you ;-). I will also get the same exposed on the `SchedulerClient`-interface.

kagkarlsson: True, lets go with only the consumer-version for now, and consider adding the list-version if there is popular demand :)

MarkvV: Updated the PR with some commits, I think I've implemented as discussed. Let me know if you have any more feedback.

src/main/java/com/github/kagkarlsson/scheduler/JdbcTaskRepository.java
```diff
@@ -92,7 +92,17 @@ public boolean createIfNotExists(Execution execution) {
 
 	@Override
 	public List<Execution> getDue(Instant now) {
-		return getDue(now, MAX_DUE_RESULTS);
+		return getDue(now, MAX_RESULTS);
+	}
+
+	@Override
+	public List<Execution> getScheduledExecutions() {
```
kagkarlsson: My thought was that these methods should also take a consumer that is called for every row in the resultset, so  `ExecutionResultSetMapper` needs to be refactored (if possible) to take a consumer instead. (`getDue` should return a `List` though, but might use the refactored `ExecutionResultSetMapper` by sending in a consumer `mylist::add`)

MarkvV: Oops, I must have forgotten about it. I updated the PR with also having the `JdbcTaskRepository` make use of the consumer. A wrapper class makes sure that the `getDue()` can still easily pass a mapper object in order to get back a list.

src/main/java/com/github/kagkarlsson/scheduler/ScheduledExecution.java
kagkarlsson: What about my comment about sending in a `T dataClass`? I think it's a bit clearer and explicit that way. (until I implement a sort of static task-descriptor)

```
void getScheduledExecutions(Consumer<ScheduledExecution<Object>> executionConsumer);

<T> void getScheduledExecutionsForTask(String taskName, T dataClass, Consumer<ScheduledExecution<T>> executionConsumer)```

Comment (kagkarlsson) [reply src/main/java/com/github/kagkarlsson/scheduler/ScheduledExecution.java]: Just some small adjustment, but I think we can merge this soon :)

Comment (MarkvV) [reply src/main/java/com/github/kagkarlsson/scheduler/ScheduledExecution.java]: `Class<T> dataClass` then surely instead of `T dataClass`?

I could then also have the data class passed to the `ScheduledExceution` where I can do some sort of "type checking", i.e.
```java
public DATA_TYPE getData() {
    if (dataClass.isInstance(this.execution.taskInstance.getData())) {
        return (DATA_TYPE) this.execution.taskInstance.getData();
    }
    
    throw new ???Exception();
}
```
Could be either a checked or unchecked exception.. do you think this would be useful?

kagkarlsson: * Yes, you are correct! `Class<T>`. A `RuntimeException` would be nice, `DataClassMismatchException` or something like that

* Could you also change the `getTaskInstance` method to return a `TaskInstanceId` instead, so that we only have a single way of accessing the data?

* And lastly, it seems reasonable to also expose the execution-time on the `ScheduledExecution`

kagkarlsson:
Looks good! Do you have any more changes, or is it ready to merge?

MarkvV:
I think this is it :). Maybe the README? But I guess that can be done whenever you release the new version.

kagkarlsson:
Ok, perfect. Yeah, I have some general README improvements to do anyway, so will have a look at it then. Hoping to release a new version soon.

kagkarlsson:
Merging

kagkarlsson:

Issue #21, opened 2018-03-19
Support for MSSQL


tomaszglinski:
I used the table definition below and it looks like it is working, about I cannot be really sure :)

```
create table scheduled_tasks (
  task_name varchar(40) not null,
  task_instance varchar(200) not null,
  task_data varchar(max),
  execution_time datetime2 not null,
  picked BIT not null,
  picked_by varchar(50),
  last_success datetime2 null,
  last_failure datetime2 null,
  last_heartbeat datetime2 null,
  version BIGINT not null,
  PRIMARY KEY (task_name, task_instance)
)
```

kagkarlsson:
If you would help me test that it works, I can add it :)

Things to keep in mind: 
* timestamp should preferably contain zone information, i.e. be an instant in time. Would `datetimeoffset` be better than `datetime2`?  
* `task_data` will hold binary data, does it work with the `varchar` type in mssql?

tomaszglinski:
Unfortunately im not an expert with MSSQL, im forced to use it:) Your remarks looks well, I will try that too . 
My data is small and worked
I didn't have issues with dates, yet it is probably so to handle clients with different zone settings? Would'n it be more efficient to store unified format like UTC?

tomaszglinski:
Tell me how i can help, i will try. Im using https://hub.docker.com/r/microsoft/mssql-server-linux/ version

kagkarlsson:
One thing you could do, if you want to be more sure of the compatibility is to run a _compatibility_ test for MSSQL. For example, modify [OracleCompatibilityTest](https://github.com/kagkarlsson/db-scheduler/blob/master/src/test/java/com/github/kagkarlsson/scheduler/compatibility/OracleCompatibilityTest.java) and run it against MSSQL (will require a jdbc-driver in pom.xml).

I would like to create a test-suite that is run against a set of docker containers, representing the common databases, but haven't been prioritized.

And regarding time zone, I think the efficiency gain would be negligible. It is much easier to just keep full _instants_ in the db.

tomaszglinski:
OK, will try. Btw, I tried to use datetimeoffset for dates, but MSSQL
silently set the type to VARCHAR...

2018-03-22 7:53 GMT+01:00 Gustav Karlsson <notifications@github.com>:

kagkarlsson:
Ok, you will have to experiment a bit on your own, I should get it covered with a integration test against a mssql container in the future..

krismorte:
#29

[...]

PR #28, opened 2018-04-20
add mssql table definition

krismorte:
I started to study the code and in my first test I was able to execute in MSSQL.
By the way the word version is reserved in MSSQL, my suggestion is that in the future it will be changed.

In the future I intend to contribute more, thanks!

### DIFF
--- a/src/test/java/com/github/kagkarlsson/scheduler/MssqlTestDatabaseRule.java
+++ b/src/test/java/com/github/kagkarlsson/scheduler/MssqlTestDatabaseRule.java
@@ -23,8 +23,8 @@ public class MssqlTestDatabaseRule extends ExternalResource {
 	public void before() throws Throwable {
 		HikariConfig config = new HikariConfig();
 		config.setJdbcUrl("jdbc:sqlserver://SERVER;databaseName=DATABASE");
-		config.setUsername("USER");
-		config.setPassword("PASSWORD");
+		config.setUsername("sa");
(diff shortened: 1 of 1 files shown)


kagkarlsson:
Thank you for the pull request! Could you run the `CompatibilityTest` against a database you have created with this schema? You probably have to add a msql jdbc-driver and create something similar to `com.github.kagkarlsson.scheduler.compatibility.OracleCompatibilityTest`.

There has also been some discussion regarding mssql schema here: #21

krismorte:
Ok, I read the issue and make some changes based on it. I also add the MssqlCompatibilityTest and MssqlTestDatabaseRule. You can drop this PR I will send another with the changes.

kagkarlsson:
Ok, great!

Btw, it looks like you have some auto-formatting going on. I would appreciate not getting any lines with only whitespace-change in the PR :)

krismorte:
Ok, but how can I fix this?
If I send another commit the blank line will going off?

kagkarlsson:
Yeah, if you revert the whitespace changes in this branch and commit, it will disappear from the changes here in the PR. 

Or you can do something like this (though in your case, a bit further than `HEAD~1`):
http://blog.ranjansakalley.com/2015/09/cleaning-up-whitespace-changes-in-a-pull-request.html

The simplest thing is perhaps to create a new branch and manually add the actual changes again. Be sure to disable auto-format on save in your IDE.

PR #29, opened 2018-04-24
add mssql table definition

krismorte:
Table script and Compatibility Test

### DIFF
--- a/src/test/resources/mssql_tables.sql
+++ b/src/test/resources/mssql_tables.sql
@@ -3,7 +3,7 @@ create table scheduled_tasks (
   task_instance varchar(250) not null,
   task_data  nvarchar(max),
   execution_time datetimeoffset  not null,
-  picked varchar(10),
+  picked bit,
   picked_by text,
   last_success datetimeoffset ,
   last_failure datetimeoffset ,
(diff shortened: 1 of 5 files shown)


src/test/resources/mssql_tables.sql
```diff
@@ -0,0 +1,13 @@
+create table scheduled_tasks (
+  task_name varchar(250) not null,
+  task_instance varchar(250) not null,
+  task_data  nvarchar(max),
+  execution_time datetimeoffset  not null,
+  picked varchar(10),
```
kagkarlsson: Isn't there a boolean or bit type in mssql?

kagkarlsson:
Looks better without the whitespace diffs, thanks :)

Did the schema work with the `CompatibilityTest`?

src/test/resources/mssql_tables.sql
krismorte: usually is used bit type

krismorte:
When I run the `CompatibilityTest` it's show me some error in the schema. So I change the table schema and run again without errors.

src/test/resources/mssql_tables.sql
kagkarlsson: If it works with `bit` that is better

krismorte: I had not looked at this particular column before. I ran the CompatibilityTest again and the result was 0, so I changed the column to bit and the result as the same. I will submit another commit.

kagkarlsson:
Ok, I will merge this and comment out the tests and the dependency (similar to what I did for Oracle). Will try to setup integration-tests against all supported datasources when I have time.

Issue #11, opened 2017-04-20
Make it possible to replace Clock in Scheduler


silmeth:
It would be nice to be able to provide another `Clock` implementation in the scheduler for testing purposes.

At the moment the `Scheduler.Builder` always passes `SystemClock` implementation which delegates `now()` to `Instant.now()`.

kagkarlsson:
Yes, I think we can do that. However, I am curious about what type of tests you are thinking about? The Scheduler will probably be hard to control from a test, as there are no hooks for forcing a check for scheduled tasks etc. So you might have to do some `Thread.sleep` ...

silmeth:
Actually I haven’t yet started writing tests involving things interacting with scheduler, so I do not have any concrete examples, and you might be right there aren’t any. I just saw that the `Scheduler` uses private `Clock`, and since we do use our own implementation where possible to be able to test time-sensitive things, I thought about putting it in the `Scheduler` too.

When / if I have anything particular on my mind, I’ll post examples here, probably somewhere in the next week.

kagkarlsson:
Ok, sure. What I have done so far for integration tests is to just leave the scheduler out of it, and call the tasks the scheduler will call directly instead. But I see how it can be useful to have integration tests that uses the scheduler, for more complete tests..

kagkarlsson:
A test-helper flavor of the scheduler, `ManualScheduler`, has been merged to master via PR #41 . It uses a clock that can be controlled from the outside.

kagkarlsson:
Fix available in `4.1`

PR #33, opened 2018-10-16
Add overridable clock

qmg-aki:
Hi there,

I had a look at #11 and it seemed quite easy to implement, so I had a go at it.

Aki

### DIFF
--- a/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
+++ b/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
@@ -392,6 +392,7 @@ public class Scheduler implements SchedulerClient {
 		private StatsRegistry statsRegistry = StatsRegistry.NOOP;
 		private Duration heartbeatInterval = Duration.ofMinutes(5);
 		private Serializer serializer = Serializer.DEFAULT_JAVA_SERIALIZER;
+        private Clock clock = new SystemClock();
 		private String tableName = JdbcTaskRepository.DEFAULT_TABLE_NAME;
 
 		public Builder(DataSource dataSource, List<Task<?>> knownTasks) {
@@ -440,6 +441,11 @@ public class Scheduler implements SchedulerClient {
 			return this;
 		}
 
+        public Builder clock(Clock clock) {
+		    this.clock = clock;
+		    return this;
+        }
+
 		public Builder tableName(String tableName) {
 			this.tableName = tableName;
 			return this;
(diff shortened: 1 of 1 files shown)


kagkarlsson:
Thanks for the contribution :)

What is your use-case, testing?

kagkarlsson:
I will have a look at the PR.

qmg-aki:
@kagkarlsson Hi, the use case is when running under a test environment an app and we have some tasks scheduled for the far future (days or even months). When we want to test those tasks it is easier to:

* Schedule the task for the future
* Forward the clock of the app at the point of the task execution
* Test that the task has executed correctly

kagkarlsson:
Ok. As I said in the #11 you may have to resort to `Thread.sleep(<pollingInterval>)` to be sure the scheduler have run at the new time. It may be a good idea to instead have a method for triggering the scheduler inline, in the current thread? (It is what I have been doing in the unit-tests)

A general comment though is that I feel configurability that supports unit/integration-testing should get a separate builder-class, so it does not "pollute" that `Scheduler.Builder`

qmg-aki:
@kagkarlsson 

> As I said in the #11 you may have to resort to Thread.sleep(<pollingInterval>)

If I understand correctly this would be due to the polling interval correct? Then that means that we just having to wait for the next polling interval to kick in, then the Scheduler will compare the current (overridden) time with the timestamps of the tasks in the database and should start executing the tasks.

> I feel configurability that supports unit/integration-testing should get a separate builder-class

I have added a separate class as you requested, let me know if it is what you were thinking of.

kagkarlsson:
Yeah, it is because the `pollingInterval`, and because everything happens in threads you do not control in the test. Though, one trick that could be used is to count down a [CountDownLatch](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CountDownLatch.html) in the `ExecutionHandler` of the task you are waiting for, and simply call `.await()` on the latch at the end of the test (with some timeout).

When I have tested similar things in the scheduler unit tests, I have used a sort of inline-version of the Scheduler, where you can manually trigger it to go and execute due executions, and everything is run synchronously in the thread running the unit test. (It also uses a `Clock` that can be changed programatically).

kagkarlsson:
Though I am not a big fan of test-code in the `main` artefact, to avoid having to create a multi-module maven project, how about creating a package `package com.github.kagkarlsson.scheduler.test;` where helpers for testing is put. 

And for starters a class 
`TestHelper` with some method that creates a `TestSchedulerBuilder`, where clock can be set and whatever else is necessary. It could subclass `Scheduler.Builder` is that helps.

kagkarlsson:
I am working on this in the branch `test-helpers`.

kagkarlsson:
Sorry for being a bit opinionated on the matter :)
I have implemented a test-helper in PR #41 which has been merged into master. 

Will close this PR.

qmg-aki:
@kagkarlsson sorry for not getting back to you, I have been away. Good to see you have that feature on master, I'm looking forward to using it on your next release.

kagkarlsson:
Just released `4.1` with this in it. Let me know it is not working according to your needs.

Issue #46, opened 2018-12-18
Flag to keep scheduler busy when there are many small one-time tasks?


bgooren:
Hi!

I'm running into an issue with low throughput of tasks when we have many one-time tasks.

My use-case is as follows: every morning we want to send a certain set of users an e-mail. As an experiment with this library, I now schedule a new one-time task instance with the user id. This is a nice design as it makes retries possible in case of temporary failure (e.g. transactional e-mail provider has temporary outage), also across restarts of the application.

In testing we now learned that (with 2 threads and a 10 second polling interval), we can send 2 emails every 10 seconds.

What I expected was that the scheduler would keep processing tasks and only start polling at 10 second intervals once there was no job to process right away.
Now it essentially runs two one-time tasks within a second and then sits there waiting for 9+ seconds.

It would make sense (to me) from a throughput perspective to poll the queue for new work when we complete, and if there is none, then we wait for the configured poll duration.
E.g. a flag like ```.pollAfterCompletion(boolean)```.

Alternatively I can lower the polling duration (e.g. to 250ms), but that means we are polling very frequently when there is nothing to be done.
Increasing the number of threads is not a solution for us as this leads to other issues (max number of simultaneous connections we can make to third party providers).

What are your thoughts on this?

kagkarlsson:
Will have to check the code, will get back to you

kagkarlsson:
Yes you are correct. I agree that the ideal behavior is for the scheduler to keep executing until there are no executions left. The current implementation is a bit naive.

Probably no quick-fix, but needs to be fixed..

bgooren:
PS Another topic when having a queue full of one-time tasks is that it's a bit wasteful to pull down up to 10.000 tasks from the database every time we poll (JdbcTaskRepository.MAX_RESULTS).
For our workload a much lower number would work as well (due to limited concurrency/competition between schedulers). Are you open to making this configurable with a builder method as well? E.g. ```pollingLimit``` (with a default of ```10000```).

bgooren:
If you are open to it, I can create pull requests.

kagkarlsson:
Please do :)

tir. 18. des. 2018 kl. 16:56 skrev Sebastian Gooren <
notifications@github.com>:

bgooren:
I have created pull requests for both; Please note I did not write additional tests to verify the behavior of the new functionality (yet).

Please have a look and provide feedback.

Thank you!

kagkarlsson:
[...]

PR #47, opened 2018-12-18
Allow configuration of polling limit

bgooren:
Fixes part of #46

### DIFF
--- a/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
+++ b/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
@@ -414,7 +414,8 @@ public class Scheduler implements SchedulerClient {
 		public Builder(DataSource dataSource, List<Task<?>> knownTasks) {
 			this.dataSource = dataSource;
 			this.knownTasks.addAll(knownTasks);
-			defaultPollingLimit();
+			this.pollingLimit = calculatePollingLimit();
+			this.useDefaultPollingLimit = true;
 		}
 
 		@SafeVarargs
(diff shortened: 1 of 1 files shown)


bgooren:
Was thinking along those lines as well, but it quickly becomes dependent on more variables.

E.g. for 2 schedulers with the same performance, polling interval, thread pool size and uniform jobs, that would work.
But suppose you have fast and slow tasks, fast and slow scheduler instances, different thread pool sizes (I made it dependent on number of available cpus for example), it quickly becomes more complex.

... and then I decided that the user probably knows best. Although a better default could be 2 * thread pool size.

kagkarlsson:
Yes I don't think it will be common that the scheduler will get a long backlog of due executions, but sure, for those few cases it might be helpful to be able to tune the size (though I have a feeling it might be a premature optimization). We can keep the setting if you feel it is helpful to you though

I think we should also change the default to 2 or 3 times nr-of-threads, would you mind adding it to this PR?

bgooren:
I have updated the PR: the builder now handles this, automatically setting the polling limit based on the number of threads, unless the polling limit is manually overriden.
When the number of threads is changed, the polling limit is recalculated.

A caller can also explicitly indicate he wants to use the default polling limit by calling ```builder.defaultPollingLimit()```, although it is the default.

I have added a static int ```POLLING_CONCURRENCY_MULTIPLIER``` (set to 3) which is used to calculate the default polling limit.

Have a look and let me know what you think.

bgooren:
Can you please merge these changes? I would like to test my other PR in production, and would like to do so with a more reasonable polling limit.

kagkarlsson:
👍

PR #48, opened 2018-12-18
Allow check for new due executions after completing picked tasks

### LINKED ISSUE #46: Flag to keep scheduler busy when there are many small one-time tasks?

bgooren:
Fixes #46

### DIFF
--- a/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
+++ b/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
@@ -46,7 +46,7 @@ public class Scheduler implements SchedulerClient {
 	private final Waiter executeDueWaiter;
 	private boolean enableImmediateExecution;
 	private final boolean pollAfterCompletion;
-	private boolean lastPollAcquiredAllExecutors;
+	private boolean lastPollAcquiredAllAvailableExecutors;
 	protected final List<OnStartup> onStartup;
 	private final Waiter detectDeadWaiter;
 	private final Duration heartbeatInterval;
@@ -188,7 +188,7 @@ public class Scheduler implements SchedulerClient {
 
 		int pickedExecutionCount = 0;
 		int availableExecutorCount = executorsSemaphore.availablePermits();
-		lastPollAcquiredAllExecutors = false;
+		lastPollAcquiredAllAvailableExecutors = false;
 		LOG.trace("Found {} taskinstances due for execution", dueExecutions.size());
 		for (Execution e : dueExecutions) {
 			if (schedulerState.isShuttingDown()) {
@@ -201,7 +201,7 @@ public class Scheduler implements SchedulerClient {
 				pickedExecution = aquireExecutorAndPickExecution(e);
 			} catch (NoAvailableExecutors ex) {
 				LOG.debug("No available executors. Skipping {} due executions.", dueExecutions.size() - pickedExecutionCount);
-				lastPollAcquiredAllExecutors = true;
+				lastPollAcquiredAllAvailableExecutors = true;
 				return;
 			}
 
@@ -215,7 +215,7 @@ public class Scheduler implements SchedulerClient {
 			pickedExecutionCount++;
 		}
 
-		lastPollAcquiredAllExecutors = pickedExecutionCount >= availableExecutorCount;
+		lastPollAcquiredAllAvailableExecutors = pickedExecutionCount >= availableExecutorCount;
 	}
 
 	private Optional<Execution> aquireExecutorAndPickExecution(Execution execution) {
@@ -246,7 +246,7 @@ public class Scheduler implements SchedulerClient {
 			statsRegistry.registerUnexpectedError();
 		}
 
-		if (pollAfterCompletion && lastPollAcquiredAllExecutors) {
+		if (pollAfterCompletion && lastPollAcquiredAllAvailableExecutors) {
 			triggerCheckForDueExecutions();
 		}
 	}
--- a/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
+++ b/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
@@ -46,6 +46,7 @@ public class Scheduler implements SchedulerClient {
 	private final Waiter executeDueWaiter;
 	private boolean enableImmediateExecution;
 	private final boolean pollAfterCompletion;
+	private boolean lastPollAcquiredAllExecutors;
 	protected final List<OnStartup> onStartup;
 	private final Waiter detectDeadWaiter;
 	private final Duration heartbeatInterval;
@@ -185,11 +186,13 @@ public class Scheduler implements SchedulerClient {
 		Instant now = clock.now();
 		List<Execution> dueExecutions = taskRepository.getDue(now);
 
-		int count = 0;
+		int pickedExecutionCount = 0;
+		int availableExecutorCount = executorsSemaphore.availablePermits();
+		lastPollAcquiredAllExecutors = false;
 		LOG.trace("Found {} taskinstances due for execution", dueExecutions.size());
 		for (Execution e : dueExecutions) {
 			if (schedulerState.isShuttingDown()) {
-				LOG.info("Scheduler has been shutdown. Skipping {} due executions.", dueExecutions.size() - count);
+				LOG.info("Scheduler has been shutdown. Skipping {} due executions.", dueExecutions.size() - pickedExecutionCount);
 				return;
 			}
 
@@ -197,7 +200,8 @@ public class Scheduler implements SchedulerClient {
 			try {
 				pickedExecution = aquireExecutorAndPickExecution(e);
 			} catch (NoAvailableExecutors ex) {
-				LOG.debug("No available executors. Skipping {} due executions.", dueExecutions.size() - count);
+				LOG.debug("No available executors. Skipping {} due executions.", dueExecutions.size() - pickedExecutionCount);
+				lastPollAcquiredAllExecutors = true;
 				return;
 			}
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
```diff
@@ -237,6 +239,10 @@ private void releaseExecutor(Execution execution) {
 			LOG.error("Released execution was not found in collection of executions currently being processed. Should never happen.");
 			statsRegistry.registerUnexpectedError();
 		}
+
+		if(pollAfterCompletion && currentlyProcessing.isEmpty()) {
```
kagkarlsson: This should work for now. 

Am thinking that ideally we need to re-write the way `executeDue` works, perhaps to a model where the threads poll blockingly and continously from a queue of executions, and some other thread populates it. But that is a lot more work..

kagkarlsson: Just a small formatting issue ;)

kagkarlsson: Is `pollAfterLastCompleted` slightly better naming-wise?

bgooren: Yeah, wasn't 100% sure about the name yet. Your suggestion is a better name.

I also agree that ```executeDue``` could be more intelligent; Although your suggestion is a nice way to design it (one thread polls database and pushes tasks to the local execute queue, producer/consumer model), it may be overkill.

And in the end, the code for the producer would essentially be the same as in ```executeDue```.

Something like this?
```
jobsDue = fetch_due_jobs();

while(jobsDue.size() > 0 && we_need_to_keep_running) {
   try_to_pick_jobs(jobsDue);

   if(local_queue_full) {
      wait_for_queue_to_have_space();
   }

   jobsDue = fetch_due_jobs();
}

// exit loop, fall back to poll interval which triggers above loop again
```

Also something to ponder: when one spot in the thread pool opens up, do you fetch, pick and run a task, or do you batch them (50% space in thread pool, 2 threads, ...), or do you wait for the thread pool to be completely done (as in my current code)?

kagkarlsson: > Also something to ponder: when one spot in the thread pool opens up, do you fetch, pick and run a task, or do you batch them (50% space in thread pool, 2 threads, ...), or do you wait for the thread pool to be completely done (as in my current code)?

Yeah this is also an aspect of the suggested solution. If you have 9 fast executions and 1 slow, the next batch will not be started after the slow one, so 9 threads will be more idle than necessary. A producer/consumer model will make that easier to fix.. Though I see what you are hinting at in your pseudo-code, hmm..

kagkarlsson: Ideally you run as soon as there is a thread available

bgooren: I think that with a sensible default polling limit it makes (more) sense to run whenever a thread is free. This will basically prevent the "producer" from sleeping much while there are more tasks that are due.

kagkarlsson:
I thinking `pollAfterCompletion(true)` should be the default behavior, I think it is a bug that is does not do it by default. The problem is though that someone might have become reliant on the current behavior, so am thinking I will in that case bump the major version and write a comment about the changed behavior in the release note/upgrade notes. Sounds ok?

Then the details of how `pollAfterCompletion(true)` works under the hood can be changed in other PRs.

src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
bgooren: > Just a small formatting issue ;)

Sorry, I'm not sure what you meant here, can you explain?

bgooren:
Sounds ok to me; I agree that this is a significant change in behavior which warrants a new major version.

I have made ```pollAfterCompletion(true)``` the default behavior.
I have also updated the behavior so it triggers a check for due executions whenever an executor is released.

kagkarlsson:
Have you tested the changes for your use-case? Works as expected?

src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
kagkarlsson: Just a cosmetic thing, missing a whitespace `if (`  ;)

src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
```diff
@@ -237,6 +239,10 @@ private void releaseExecutor(Execution execution) {
 			LOG.error("Released execution was not found in collection of executions currently being processed. Should never happen.");
 			statsRegistry.registerUnexpectedError();
 		}
+
+		if(pollAfterCompletion) {
```
kagkarlsson: Just a thought, we might also check here if we got a full batch of executions, if not there is not need to check again

bgooren:
I have added a ```lastPollAcquiredAllAvailableExecutors``` flag which tracks if the last poll filled the available queue. We check it in ```releaseExecutor()```.

src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
```diff
@@ -46,6 +46,7 @@
 	private final Waiter executeDueWaiter;
 	private boolean enableImmediateExecution;
 	private final boolean pollAfterCompletion;
+	private boolean lastPollAcquiredAllAvailableExecutors;
```
kagkarlsson: Could perhaps be made local to the method `executeDue()`? Might have to change to an AtomicBoolean in that case

bgooren:
> Have you tested the changes for your use-case? Works as expected?

I'm in the process of adding the git repo to our internal CI so I can test a snapshot in more detail. Will keep you posted.

src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
bgooren: Sorry - I don't follow. Local as in: a local variable in the ```executeDue``` method?

kagkarlsson: Yeah, sorry wrote it in a hurry. Declare the variable in the method in stead of in a field

bgooren: Since we need to track some data between polls (whether the last poll filled up our local work queue), I'm not sure what you suggest is possible.

Can you give me some (pseudo) code so I can attempt to rework this PR?

kagkarlsson: Hmm, I did not realize we need to track the data between polls, what is the reason for that?

I was thinking to declare `lastPollAcquiredAllAvailableExecutors` in the method, something like this:
```
protected void executeDue() {
  ...
  AtomicBoolean lastPollAcquiredAllAvailableExecutors = false;
  ...
          .thenRun(() -> releaseExecutor(pickedExecution.get(), lastPollAcquiredAllAvailableExecutors));
  ...
```

src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
```diff
@@ -208,8 +214,10 @@ protected void executeDue() {
 			} else {
 				LOG.debug("Execution picked by another scheduler. Continuing to next due execution.");
 			}
-			count++;
+			pickedExecutionCount++;
 		}
+
+		lastPollAcquiredAllAvailableExecutors = pickedExecutionCount >= availableExecutorCount;
```
kagkarlsson: Was looking at this line and was wondering if that is really needed, is it not sufficient to just check for `NoAvailableExecutors` exception?

Also, `pickedExecutionCount` will include executions that was picked by another Scheduler-instance, i.e. not running in this executor-pool

bgooren: Yeah, you are correct; Checking for `NoAvailableExecutors` should be enough.

My reason for including the additional logic was an edge case; But it should be quite rare so I will simplify the code 👍

bgooren:
> > Have you tested the changes for your use-case? Works as expected?
> 
> I'm in the process of adding the git repo to our internal CI so I can test a snapshot in more detail. Will keep you posted.

The code is running in production now - I will get back to you in a few days with an update after monitoring how it runs.

kagkarlsson:
Sounds good 👍

bgooren:
Looks good so far - throughput has increased dramatically 👍 

Before:
`pollingInterval` set to 1 second, 4 threads
4 tasks (e-mails sent) per second

After:
`pollingInterval` set to 1 minute, 4 threads
`pollAfterCompletion` enabled
54 tasks (e-mails sent) per second

kagkarlsson:
Good to hear!

kagkarlsson:
Let me know when you are happy with the testing and ready for the PR to be merged

bgooren:
Works like a charm in production, so feel free to merge my PR.

Issue #55, opened 2019-03-13
Spring / SpringBoot integration


pnowy:
Hi Gustav,

first of all nice work!

Did you consider to prepare separate module (ie. db-scheduler-spring) which provide seamless integration with Spring/Spring Boot?

I was thinking about that a little bit and I think that kind of option could help library gain more popularity. From my perspective what needs to be done to do that:

- prepare Spring Boot autostarter which creates the table and initialize all required beans
- provide the option to to declare task to execute as Spring managed beans so executed task will be retrieved from context and run (in order to that task should implement specific task interface)
- the cron jobs support would be also good for that (but I've seen task for it)

Are you open to that integration? What do you think about that?

kagkarlsson:
Hi!

Sure I think that sounds like a good idea provided it does not require too many changes to the core db-scheduler code. I have so far tried to avoid a multi-module project, but this might require it, unless we maintain it in a separate project `db-scheduler-addons` (or something like that)

kagkarlsson:
Is this something you would be willing to contribute to?

pnowy:
Yes, that's my plan but I was thinking about separate module for spring - not something like addons. From my perspective it works quite good. I have similar project and using there that approach: https://github.com/pnowy/NativeCriteria

I have some core and spring if someone wants to use direct integration (but I'm also not a big fan of multimodules projects but less than 5 it's ok for me).

kagkarlsson:
Yeah multi-module is probably the best option. Slighty painful to have to go for a multi-module project structure (increased complexity), but I see no way around it if we want to offer different "flavors" of the lib..

kagkarlsson:
So either

```
db-scheduler/
    db-scheduler-core/
    db-scheduler-spring/
```

or 

```
db-scheduler/
    core/
    spring/
```

pnowy:
I would prefer first option. 

The official spring project (https://github.com/spring-projects/spring-framework) has that library part on module name like `spring-something` so I think `db-scheduler-core`, etc. will be ok (I always trying to copy patterns from good and popular projects).

kagkarlsson:
You make a good point :). Not super-happy with the `core` part, so maybe like [guava](https://github.com/google/guava)

```
db-scheduler/
    db-scheduler/       (for better main-dependency naming)
    db-scheduler-spring/
```

kagkarlsson:
Anyway, that can be tuned later. Would be super-happy for contributions!

[...]

PR #63, opened 2019-07-23
Spring boot support

evenh:
This PR addresses #55 and provides an initial implementation for Spring Boot integration.


kagkarlsson:
Awesome, thanks! Will have a look! :)

evenh:
Please feel free to be frank in the review. I know Spring Boot better than the DB scheduler :-)

db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/autoconfigure/DbSchedulerAutoConfiguration.java
```diff
@@ -0,0 +1,98 @@
...
+    @ConditionalOnBean(DataSource.class)
+    @ConditionalOnMissingBean
+    @Bean(initMethod = "start", destroyMethod = "stop")
+    public Scheduler scheduler(DbSchedulerCustomizer customizer) {
+        log.info("Creating DB Scheduler using tasks from Spring context: {}", configuredTasks);
+
+        if (existingDataSource instanceof TransactionAwareDataSourceProxy) {
+            log.info("Using a transaction aware DataSource");
+        } else {
+            log.info("The configured DataSource is not transaction aware: {}", existingDataSource);
```
kagkarlsson: Maybe we should say something about the consequences here, i.e. operations such as `schedule`, `reschedule`, `cancel` will have effect regardless of spanning transaction

db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/config/DbSchedulerProperties.java
```diff
@@ -0,0 +1,121 @@
...
+
+    public String getTableName() {
+        return tableName;
+    }
+
+    public void setTableName(final String tableName) {
+        this.tableName = tableName;
+    }
+
+    public boolean isImmediateExecution() {
```
kagkarlsson: Suffix with `isImmediateExecutionEnabled`?  (same for setter)

db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/autoconfigure/DbSchedulerAutoConfiguration.java
```diff
@@ -0,0 +1,98 @@
...
+@ConditionalOnBean(DataSource.class)
+public class DbSchedulerAutoConfiguration {
+    private static final Logger log = LoggerFactory.getLogger(DbSchedulerAutoConfiguration.class);
+
+    private final DbSchedulerProperties config;
+    private final DataSource existingDataSource;
+    private final List<Task<?>> configuredTasks;
+
+    public DbSchedulerAutoConfiguration(DbSchedulerProperties dbSchedulerProperties,
+        DataSource dataSource, List<Task<?>> configuredTasks) {
```
kagkarlsson: Will this inject all instances of `Task` on the classpath?

kagkarlsson (commented):
Looks good, though probably will need a second scan of things :)

Would be neat with some short example `Main` showing a minimal example, ala for example https://github.com/kagkarlsson/db-scheduler/blob/master/src/test/java/com/github/kagkarlsson/scheduler/example/CronMain.java

kagkarlsson:
Hmm, I think we are missing the `startTasks(...)` aspect for kicking off initial execution for recurring tasks, see https://github.com/kagkarlsson/db-scheduler/blob/master/src/test/java/com/github/kagkarlsson/scheduler/example/TasksMain.java#L41

Tasks that also implement `OnStartup` should be added to the builder via `startTasks(...)` method

db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/autoconfigure/DbSchedulerAutoConfiguration.java
evenh: It should inject all `Task`-instances in the Spring context, or an empty list if no `Task` beans is defined.

Should we throw an exception if the list of configured `Task` beans is empty or just let the scheduler idle away?

evenh:
I've implemented the `startTasks` invocation in the builder, where I've filtered out tasks which do and do not implement `OnStartup`.

As for the examples, should they be separate "real" Spring Boot projects? I was thinking something like (Maven hierarchy):

```
| db-scheduler
| db-scheduler-boot-starter
| examples
|  --> spring-boot-example
|  --> vanilla-java-example?
|  --> other-example?
```

The benefit of this approach is breakage of the examples if the library/integrations change :-)

db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/autoconfigure/DbSchedulerAutoConfiguration.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    @ConditionalOnBean(DataSource.class)
+    @ConditionalOnMissingBean
+    @Bean(initMethod = "start", destroyMethod = "stop")
+    public Scheduler scheduler(DbSchedulerCustomizer customizer) {
+        log.info("Creating DB Scheduler using tasks from Spring context: {}", configuredTasks);
+
+        if (existingDataSource instanceof TransactionAwareDataSourceProxy) {
+            log.info("Using a transaction aware DataSource");
```
cash1981: Is this really necessary information to give? I guess the else block info makes more sense than this?

evenh: Probably not worthy of an `info` but a `debug` might be handy?

db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/autoconfigure/DbSchedulerAutoConfiguration.java
```diff
@@ -0,0 +1,131 @@
...
+        } else {
+            log.info("The configured DataSource is not transaction aware: '{}'. Operations such as schedule, reschedule and cancel will have effect regardless of spanning transactions.", existingDataSource);
```
cash1981: This log statement is a bit confusing. What is the purpose of this?

kagkarlsson: Are there cases where a Spring-user might send in a datasource is not transaction-aware by mistake?

cash1981:
Perhaps also update the documentation on how to add db-scheduler using Spring and this PR?

db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/autoconfigure/DbSchedulerAutoConfiguration.java
evenh: @cash1981: I thought that information whether one uses an instance of `TransactionAwareDataSourceProxy` would be useful information, see the [docs](https://github.com/kagkarlsson/db-scheduler#things-to-note--gotchas).

However, I can't seem to get Spring Boot's configuration to provide me with a `TransactionAwareDataSourceProxy` out of the box (using JPA or JDBC starter, `@EnableTransactionManagement` on/off and using a local PostgreSQL db).

evenh:
Absolutely @cash1981! As pointed out [in this comment](https://github.com/kagkarlsson/db-scheduler/pull/63#issuecomment-514825576), we should consider adding some examples and link to those as well.

kagkarlsson:
@evenh Yeah for spring-boot I think it makes sense to have an examples-directory with a full project that is part of the build. We can link to it from the documentation. If necessary, code can be  copied to the markdown-doc. Maybe a separate markdown-page for usage in spring-boot which we link to. 

If you could create an example project @evenh , I can update the doc :)

evenh:
Fixed @kagkarlsson :)

kagkarlsson:
Ran the example and shutdown seem to be working as exepected @cash1981

kagkarlsson:
I think we need to wrap the datasource in a `TransactionAwareDataSourceProxy` if it is not already

kagkarlsson:
Enabled maven-dependency-plugin:

```[INFO] --- maven-dependency-plugin:2.10:analyze-only (default) @ db-scheduler-spring-boot-starter ---
[WARNING] Used undeclared dependencies found:
[WARNING]    org.springframework.boot:spring-boot-test:jar:2.1.6.RELEASE:test
[WARNING]    org.assertj:assertj-core:jar:3.11.1:test
[WARNING]    org.springframework.boot:spring-boot-autoconfigure:jar:2.1.6.RELEASE:compile
[WARNING]    org.springframework.boot:spring-boot-actuator:jar:2.1.6.RELEASE:compile
[WARNING]    org.springframework.boot:spring-boot:jar:2.1.6.RELEASE:compile
[WARNING]    junit:junit:jar:4.12:test
[WARNING]    org.slf4j:slf4j-api:jar:1.7.26:compile
[WARNING]    org.springframework:spring-context:jar:5.1.8.RELEASE:compile
[WARNING] Unused declared dependencies found:
[WARNING]    org.springframework.boot:spring-boot-configuration-processor:jar:2.1.6.RELEASE:compile
[WARNING]    org.springframework.boot:spring-boot-starter:jar:2.1.6.RELEASE:compile
[WARNING]    org.springframework.boot:spring-boot-autoconfigure-processor:jar:2.1.6.RELEASE:compile```

Will see if I can fix a couple..

kagkarlsson:
Merging this PR. Further improvements can be suggested as separate issues.

kagkarlsson:

PR #65, opened 2019-08-19
Fix dead links to table definitions

runeflobakk:
Links stopped working after restructuring to db-scheduler core module. Change to relative paths instead of full URLs.


kagkarlsson:
Ah, I didn't notice that, thanks alot!! :)     🏅

PR #68, opened 2019-09-20
Support time zones in CronExpression

paulhilliar:
Support time zones in CronExpression so that schedules can fire at a time of day that moves specific to the explicit zone not the zone of the JVM.

For example the expression "0 05 13,20 * * ?" (New York) should fire at 13:05/20:05 NY time every day, regardless of where the JVM sits.


kagkarlsson:
Excellent PR! Thanks for contributing!

Could you add a `cron(String, ZoneId)` method to the `Schedules` class as well?

paulhilliar:
Thanks.  Added the Schedules method.

kagkarlsson:
Thanks! Merging!

kagkarlsson:
Should be available in maven central in version 6.1 soon

kagkarlsson:

PR #71, opened 2019-10-14
Configurable scheduler start

### LINKED ISSUE #70: Scheduler start in Spring-Boot
Hi Gustav, 
i have the problem in a Spring-Boot application, that the scheduler starts before the spring context loading is finished. Is there any out of the box way to start the scheduler threads only **after** the context is initialized? 
(The only way I found, is to rewrite the autoconfiguration using the ManualSchedulerBuilder as Builder, overwriting the _executeOnStartup_ method).

Is there any better/simple way to do this? No one had this issue before?

Thanks.

evenh:
This PR resolves #70 by introducing a new configuration property: `db-scheduler.start-scheduler-immediately`.

Existing behaviour is enabled by default (`true`). However, if it is set to `false` the scheduler will start after the Spring context is ready/refreshed ([`ContextRefreshedEvent`](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/context/event/ContextRefreshedEvent.html)).

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -82,7 +82,9 @@ This software includes third party software subject to the following licenses:
   Spring Boot Actuator AutoConfigure under Apache License, Version 2.0
   Spring Boot Actuator Starter under Apache License, Version 2.0
   Spring Boot AOP Starter under Apache License, Version 2.0
+  Spring Boot Auto-Configure Annotation Processor under Apache License, Version 2.0
   Spring Boot AutoConfigure under Apache License, Version 2.0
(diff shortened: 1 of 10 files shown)


kagkarlsson:
Thanks!

How do you feel about calling the property `delay-startup-until-context-ready`, or `start-after-context-ready`? (or something like that)

evenh:
Does this solve your problem @zsmeier?

zsmeier:
Looks good! Thank you!!!!

PR #82, opened 2020-01-13
move junit explicitly to the test scope

ystarikovich:


kagkarlsson:
Oh, good catch. Missed this one. Thanks 👍

kagkarlsson:
Do you need a new release with this asap?

ystarikovich:
Not super urgent right now. I'm also thinking to submit more generic Customizer which we used in our auto configuration on behalf of @poberbichler. What do you think about this enhancement, then Scheduler bean creation should be simplified.

```

import com.github.kagkarlsson.scheduler.SchedulerBuilder;

/**
 * Callback interface that can be used to customize a {@link SchedulerBuilder}
 */
@FunctionalInterface
public interface DbSchedulerCustomizer {
    void customize(SchedulerBuilder schedulerBuilder);
}
```

kagkarlsson:
It was actually released yesterday since I had another fix I needed released.

Will have a look at your suggestion for spring boot

kagkarlsson:
Your suggestion for `DbSchedulerCustomizer` makes sense. There might be a source of confusion since you will be able to override properties from `DbSchedulerProperties`, but if it is mentioned explicitly in the javadoc that should be fine.

Customizer apply probably should be moved to the end of `DbSchedulerAutoConfiguration.schedule(..)` method, so that anything you set in the customizer will have precedence.

kagkarlsson:
Feel free to submit a PR :)

PR #85, opened 2020-01-21
Add @ConditionalOnEnabledHealthIndicator

ystarikovich:
To follow Spring best practices and give a possibility to disable/enable health indicator @ConditionalOnEnabledHealthIndicator was introduced

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ See also [why not Quartz?](#why-db-scheduler-when-there-is-quartz)
 </dependency>
 ```
 
-2. Create the `scheduled_tasks` table in your database-schema. See table definition for [postgresql](db-scheduler/src/test/resources/postgresql_tables.sql), [oracle](db-scheduler/src/test/resources/oracle_tables.sql), [mssql](db-scheduler/src/test/resources/mssql_tables.sql) or [mysql](db-scheduler/src/test/resources/mysql_tables.sql).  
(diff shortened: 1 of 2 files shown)


kagkarlsson:
Excellent! I am unfortunately not very familiar with Spring Boot best practices, can I find them somewhere?

So with this change you have to "opt-in" to get a `HealthIndicator`? 
Are `HealthIndicators` primarily used to indicate the healthiness of the application, i.e. if it should receive traffic?

ystarikovich:
I made a link to the spring documentation, there is information how HealthIndicators are living in the Spring ecosystem and also how anyone can configure behaviour of particular indicator or all indicators in general.

There is no possibility to control this HealthCheck behaviour right now. From perspective of our application and maybe someone else will have the same problem, db-scheduler availability should not be a indicator for container orchestration framework to kill instance of application as this is non critical part and application should receive traffic if something happened with db-scheduler.

kagkarlsson:
I agree that if the `HealthIndicator` is used to signal the health/readiness of the application it should default to off, since db-scheduler is really a sort of independent sub-system living in the same JVM. We can serve user-traffic without it.

kagkarlsson:
Let me know if you need a release with this fix.

ystarikovich:
Would appreciate if you can release this as right now we are using workaround for this.

kagkarlsson:
Np. Release 6.6 should be available in central now 👍

PR #87, opened 2020-02-05
Support timezone in daily schedule

alex859:
Hello! First of all thank you very much for this project, very useful!

Similarly to #68, it would be also useful to support time zones in `Daily` schedules.


kagkarlsson:
Excellent! Thank you for contributing! :) 

Will have a look at the PR as soon as I can. Looks good on first glance.

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/task/DailyTest.java
```diff
@@ -5,47 +5,138 @@
...
+        assertThat(new Daily(ROME, HOUR_2).getNextExecutionTime(complete(instant(currentDate, HOUR_0, UTC))), is(instant(currentDate, HOUR_1, UTC)));
+        assertThat(new Daily(ROME, HOUR_3).getNextExecutionTime(complete(instant(currentDate, HOUR_0, UTC))), is(instant(currentDate, HOUR_1, UTC)));
+    }
+
+    @Test
+    public void should_generate_next_date_correctly_on_the_day_when_time_changes_to_standard() {
+        // Europe/Rome is UTC+1 during standard time
+        // Europe/Rome is UTC+2 during daylight saving time
+        // Standard saving begins on 25 October 2020 at 3AM
+        LocalDate currentDate = LocalDate.of(2020, 3, 29);
```
kagkarlsson: Did you mean to use october here?

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/task/DailyTest.java
```diff
@@ -5,47 +5,138 @@
...
+    @Test
+    public void should_generate_next_date_correctly_during_standard_time() {
+        // Europe/Rome is UTC+1 during standard time
+        LocalDate currentDate = LocalDate.of(2020, 12, 12);
+        LocalDate nextDay = currentDate.plusDays(1);
+
+        assertThat(new Daily(ROME, HOUR_8).getNextExecutionTime(complete(instant(currentDate, HOUR_0, UTC))), is(instant(currentDate, HOUR_7, UTC)));
+        assertThat(new Daily(ROME, HOUR_8).getNextExecutionTime(complete(instant(currentDate, HOUR_7, UTC))), is(instant(nextDay, HOUR_7, UTC)));
+    }
+
```
kagkarlsson: Good job adding all the tests for the edge-cases 👍

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/task/DailyTest.java
alex859: Yup you are absolutely right :relaxed:, fixed.

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/task/DailyTest.java
alex859: Thank you!

alex859:
What would you think about supporting a daily string format like: `DAILY|Timezone|hh:mm,hh:mm,...,hh:mm` ?

kagkarlsson:
> What would you think about supporting a daily string format like: `DAILY|Timezone|hh:mm,hh:mm,...,hh:mm` ?

Yeah, that sounds like a good idea. I added the string-based formats to easier enable configuring schedules from a config-file.

alex859:
Cool, should I do that as part of this PR?

kagkarlsson:
Yeah you can just add it here 👍

alex859:
Hello @kagkarlsson!

I've added the timezone to the daily string format as `DAILY|13:30...|Timezone`.

I took this opportunity to refactor the parsing logic. This could be useful if more parsers were required  in future.

Please let me know your thoughts :relaxed:.

kagkarlsson:
Great! Will have a look as soon as possible

kagkarlsson:
Do you have a time when you need it released?

alex859:
Thank you! 

It's not urgent, we'll replace `cron` expressions in our config files whenever this is available.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/DailyParser.java
```diff
@@ -0,0 +1,87 @@
...
+            .orElse(new Daily(matchedSchedule.scheduleTimes()));
+    }
+
+    private MatchedSchedule match(String scheduleString) {
+        return Optional.ofNullable(scheduleString)
+            .map(DAILY_PATTERN_WITH_TIMEZONE::matcher)
+            .filter(Matcher::matches)
+            .map(MatchedSchedule::new)
+            .orElseThrow(() -> new UnrecognizableSchedule(scheduleString, this.examples()));
+    }
```
kagkarlsson: Have you considered making an abstract parent class that might dry things up even more? Have a feeling all parsers will be regex-based..

```
public abstract class RegexParser implements Parser {
....
```

README.md
```diff
@@ -160,10 +160,18 @@ The library contains a number of Schedule-implementations for recurring tasks. S
...
 | `.cron(String)`  | Spring-style cron-expression. |
 
+Another option to configure schedules is reading string patterns with `Schedules.parse(String)`.
+
+The currently available patterns are:
+
+| Pattern  | Description |
+| ------------- | ------------- |
+| `FIXED_DELAY\|Ns`  | Same as `.fixedDelay(Duration)` with duration set to N seconds. |
+| `DAILY\|12:30,15:30...(\|time_zone)`  | Same as `.daily(LocalTime)` with optional time zone.|
```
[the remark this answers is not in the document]
alex859: That's right, added :+1:

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/DailyParser.java
```diff
@@ -0,0 +1,87 @@
...
+
+        List<LocalTime> scheduleTimes() {
+            String[] times = matcher.group(1).split(",");
+            return Stream.of(times).map(timeStr -> {
+                String[] hourAndMinute = timeStr.split(":");
+                return LocalTime.of(Integer.parseInt(hourAndMinute[0]), Integer.parseInt(hourAndMinute[1]));
+            }).collect(Collectors.toList());
+        }
+
+        private static Function<String, ZoneId> toZoneId() {
```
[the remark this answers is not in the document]
alex859: I found `.map(toZoneId())` slightly more readable than `.map(parseZoneId())`, what do you think?

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/FixedDelay.java
```diff
@@ -62,4 +63,22 @@ public Instant getNextExecutionTime(ExecutionComplete executionComplete) {
...
+        if (this == o) return true;
+        if (!(o instanceof FixedDelay)) return false;
+        FixedDelay that = (FixedDelay) o;
+        return Objects.equals(this.duration, that.duration);
+    }
+
+    @Override
+    public final int hashCode() {
+        return Objects.hash(duration);
+    }
```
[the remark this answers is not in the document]
alex859: Yes, I tried using it but I got some build issues after pushing..I'll try again.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/Parser.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.kagkarlsson.scheduler.task.schedule;
+
+import java.util.List;
+import java.util.Optional;
+
+interface Parser {
+    Schedule parse(String scheduleString);
```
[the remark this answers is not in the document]
alex859: True, changed :+1:

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/Schedules.java
```diff
@@ -19,63 +19,49 @@
...
-
-		}
-
-		Matcher fixedDelayMatcher = FIXED_DELAY_PATTERN.matcher(scheduleString);
-		if (fixedDelayMatcher.matches()) {
-			int secondsDelay = Integer.parseInt(fixedDelayMatcher.group(1));
-			return FixedDelay.of(Duration.ofSeconds(secondsDelay));
-		}
-
-		throw new UnrecognizableSchedule(scheduleString);
```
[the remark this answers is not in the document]
alex859: Thanks! :relaxed:

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/DailyParser.java
alex859: I've tried extracting the shared regex related logic to a `OptionalMatcher` class, what do you think?

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/CompositeParser.java
```diff
@@ -0,0 +1,51 @@
...
+    private final List<Parser> delegates;
+
+    private CompositeParser(List<Parser> delegates) {
+        this.delegates = delegates;
+    }
+
+    @Override
+    public Optional<Schedule> parse(String scheduleString) {
+        return delegates.stream()
+            .map(NotThrowingParser::of)
```
alex859: Not sure if this is required, I thought we should still guard against parser implementations that throw exceptions.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/FixedDelay.java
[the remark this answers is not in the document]
alex859: Build still fails, looks like the new `NOTICE` file doesn't match yours? :man_shrugging:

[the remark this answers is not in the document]
kagkarlsson: Yeah I should probably clean up the NOTICE-stuff. Doesn't really make sense to have test-dependencies appear there..  But for now you can just add it so it matches.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/DailyParser.java
kagkarlsson: I still feel that the code might be clearer if you structure the code similar to this:

```
public abstract class RegexBasedParser implements Parser {
    private final Pattern pattern;
    private final List<String> examples;

    public RegexBasedParser(Pattern pattern, List<String> examples) {
        this.pattern = pattern;
        this.examples = examples;
    }

    @Override
    public Optional<Schedule> parse(String scheduleString) {
        Matcher m = pattern.matcher(scheduleString);
        if (m.matches()) {
            return getMatchedSchedule(m);
        } else {
            return Optional.empty();
        }
    }

    @Override
    public List<String> examples() {
        return examples;
    }

    public abstract Optional<Schedule> getMatchedSchedule(MatchResult result);

}

...
(example re-implementation of FixedDelayParser:)

final class FixedDelay2Parser extends RegexBasedParser {
    private static final Pattern FIXED_DELAY_PATTERN = Pattern.compile("^FIXED_DELAY\\|(\\d+)s$");
    private static final List<String> EXAMPLES = Collections.singletonList("FIXED_DELAY|120s");

    FixedDelay2Parser() {
        super(FIXED_DELAY_PATTERN, EXAMPLES);
    }

    @Override
    public Optional<Schedule> getMatchedSchedule(MatchResult result) {
        FixedDelay schedule = FixedDelay.ofSeconds(Integer.parseInt(result.group(1)));
        return Optional.of(schedule);
    }
}
```

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/DailyParser.java
```diff
@@ -0,0 +1,95 @@
...
+
+        Optional<Schedule> dailyWithTimezone() {
+            return maybeZoneIdStr().flatMap(this::maybeCorrectZoneId).map(zoneId -> new Daily(zoneId, scheduleTimes()));
+        }
+
+        Optional<Schedule> dailyWithoutTimezone() {
+            return Optional.of(new Daily(scheduleTimes()));
+        }
+    }
+}
```
kagkarlsson: To be honest, for me personally the extended use of streams and splitting the logic into multiple tiny methods makes it harder to read and mentally parse..

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/DailyParser.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        private Optional<ZoneId> maybeCorrectZoneId(String str) {
+            try {
+                return Optional.of(ZoneId.of(str));
+            } catch (DateTimeException exception) {
+                LOG.warn("Unable to parse ZoneId {}", str, exception);
+                return Optional.empty();
+            }
+        }
```
kagkarlsson: Not sure if I may be contradicting something I have said previously, but I feel the parser should be allowed to throw the exception here, instead of falling back to a default time zone. If the user had a typo in the property a fast fail is better than silently disregarding the zone :)

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/DailyParser.java
alex859: Should be clearer now :relaxed:

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/DailyParser.java
alex859: :+1:

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/FixedDelay.java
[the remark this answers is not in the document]
alex859: It should be fine now :crossed_fingers:

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/DailyParser.java
alex859: It should look clearer now :relieved:

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/DailyParser.java
alex859: :+1:

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/DailyParser.java
alex859: Yes, I agree. If the regex matches but there is something else wrong with the provided schedule, it's better to let the user know. 

Actually I reckon that `RegexBasedParser`'s abstract method should just return a `Schedule` or throw an exception (instead of returning an `Optional`).

kagkarlsson: Agreed 👍

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/RegexBasedParser.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+abstract class RegexBasedParser implements Parser {
+    private final Pattern pattern;
+    private final List<String> examples;
+
+    public RegexBasedParser(Pattern pattern, List<String> examples) {
+        this.pattern = Objects.requireNonNull(pattern);
+        this.examples = new ArrayList<>(examples);
```
kagkarlsson: Nitpicking, but unnecessary recreate of List? (just assign from parameter)

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/CompositeParser.java
kagkarlsson: As we discussed elsewhere, maybe the strategy should be:
* If the regex does not match => Optional.empty()
* If the regex matches, but some extracted field on wrong format => runtimeexception

And in this case, maybe the best thing is to let it be uncaught so it hits the user, i.e. we do not need the `NotThrowingParser`. Do you agree?

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/RegexBasedParser.java
alex859: Just to be safe from external modifications of the list :)

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/CompositeParser.java
alex859: Yes, let's do this :+1:

alex859:
I've also experienced this build failure locally, has it happened to you before?

kagkarlsson:
Sometimes I have one of the concurrent tests fail, yeah. Have fixed some problems, but maybe not all?

Let me know when you are happy with the PR being merged

alex859:
I'm happy with it, thanks!

kagkarlsson:
Heads-up. Tried to build a release earlier today, but one of the tests seemed to fail for my time-zone. I need to look at that before I can build the release

alex859:
Interesting...I can try to reproduce it, what time zone are you in?

kagkarlsson:
Oslo 🤔

tir. 25. feb. 2020 kl. 16:43 skrev Alessandro Ciccimarra <
notifications@github.com>:

kagkarlsson:
SchedulesTest

tir. 25. feb. 2020 kl. 16:56 skrev Gustav Karlsson <kagkarlsson@gmail.com>:

alex859:
My bad, Fix in #91 :relaxed:

kagkarlsson:
Release 6.7 should be available now. Thanks for the PRs!

alex859:
Thank you, my pleasure!

Issue #88, opened 2020-02-06
Make Scheduler SHUTDOWN_WAIT configurable


alexvr:
Hi Gustav,

Thanks for building db-scheduler! It has been a pleasure implementing it in our system.

I would like to make a suggestion: 

The `SHUTDOWN_WAIT` constant in the [Scheduler class](https://github.com/kagkarlsson/db-scheduler/blob/19cc6d9c9373abbe21140de435473f62c61bd5dd/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java#L40) defines that it will wait up to 30 minutes before it terminates the remaining tasks. It would be nice to make this duration configurable, so that it better suits the nature of the tasks.

I would like to hear your thoughts on this.

Regards, Alexander.

kagkarlsson:
Hi! Thank you for your kind words :)

If I remember correctly, if `SHUTDOWN_WAIT` is reached, it means some execution will be aborted. That is not ideal as it will not be "unlocked" and need to be detected as "dead" before it is re-scheduled. Usually not a big problem though..

May I suggest that you check `executionContext.getSchedulerState().isShuttingDown()` and gracefully stop whatever the task is doing when a shutdown as been detected? 

Or is your problem that you really need the scheduler to run for longer than 30min?

kagkarlsson:
@alexvr Do you still have a need to configurable `SHUTDOWN_WAIT´? (taking into consideration my comment above)

kagkarlsson:
Fixed, see builder method `.shutdownMaxWait(Duration)`

PR #92, opened 2020-02-27
Allow for configurable timeout of the Scheduler.

dpratt:
- Modify Scheduler to take responsibility for the lifecycle of
  the job ExecutorService only if it's not externally supplied.
- Make the shutdown timeout be externally configurable by the user.
- Additional logging for when the Scheduler does not shut down cleanly.
- Add bits to the POM that allow for building under Java 9+
- Have the tests use logback instead of slf4-simple for log output

This addresses #88

### DIFF
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/ExecutorUtils.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/ExecutorUtils.java
@@ -48,23 +48,34 @@ public class ExecutorUtils {
 	}
 
 	public static ThreadFactory defaultThreadFactoryWithPrefix(String prefix) {
-		return new PrefixingDefaultThreadFactory(prefix);
+		return new PrefixingDefaultThreadFactory(prefix, false);
+	}
+
+	public static ThreadFactory defaultThreadFactoryWithPrefix(String prefix, boolean useDaemonThreads) {
+		return new PrefixingDefaultThreadFactory(prefix, useDaemonThreads);
 	}
 
 	private static class PrefixingDefaultThreadFactory implements ThreadFactory {
 
 		private final String prefix;
+		private final boolean useDaemonThreads;
 		private final ThreadFactory defaultThreadFactory;
 
 		public PrefixingDefaultThreadFactory(String prefix) {
+			this(prefix, false);
+		}
+
+		public PrefixingDefaultThreadFactory(String prefix, boolean useDaemonThreads) {
 			this.defaultThreadFactory = Executors.defaultThreadFactory();
 			this.prefix = prefix;
+			this.useDaemonThreads = useDaemonThreads;
 		}
 
 		@Override
 		public Thread newThread(Runnable r) {
 			final Thread thread = defaultThreadFactory.newThread(r);
 			thread.setName(prefix + thread.getName());
+			thread.setDaemon(useDaemonThreads);
 			return thread;
 		}
 	}
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
@@ -26,8 +26,7 @@ import javax.sql.DataSource;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.*;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
+import java.util.concurrent.*;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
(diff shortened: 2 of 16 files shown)


kagkarlsson:
Thanks for the contribution! Will try and have a look in the coming days :)

kagkarlsson:
Bit of a busy week, haven't had the time to check this yet

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
```diff
@@ -26,348 +26,400 @@
 import java.time.Duration;
 import java.time.Instant;
 import java.util.*;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
+import java.util.concurrent.*;
```
kagkarlsson: Yeah I see I have a bit of a whitespace issue in the project. I will fix that afterwards, in another PR, to make it more consistent.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
```diff
@@ -26,348 +26,400 @@
...
+
+    ThreadPoolExecutor housekeepingExecutor = newThreadPool(THREAD_PREFIX + "-housekeeper-", 3, THREAD_POOL_TIMEOUT);
+    this.dueExecutor = housekeepingExecutor;
+    this.detectDeadExecutor = housekeepingExecutor;
+    this.updateHeartbeatExecutor = housekeepingExecutor;
+
+//    this.dueExecutor = Executors.newSingleThreadExecutor(defaultThreadFactoryWithPrefix(THREAD_PREFIX + "-execute-due-"));
+//    this.detectDeadExecutor = Executors.newSingleThreadExecutor(defaultThreadFactoryWithPrefix(THREAD_PREFIX + "-detect-dead-"));
+//    this.updateHeartbeatExecutor = Executors.newSingleThreadExecutor(defaultThreadFactoryWithPrefix(THREAD_PREFIX + "-update-heartbeat-"));
+
```
kagkarlsson: To be removed

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
```diff
@@ -26,348 +26,400 @@
...
+			  if(!ExecutorUtils.shutdownNowAndAwaitTermination(executorService, Duration.ofSeconds(5))) {
+			    LOG.warn("Scheduler did not terminate within the timeout. The Scheduler pool may be orphaned and have live threads.");
+        }
+        if(!currentlyProcessing.isEmpty()) {
+          String staleJobs = currentlyProcessing.keySet().stream().map(Execution::toString).collect(Collectors.joining("\n"));
+          LOG.warn("Scheduler stopped, but the following jobs were orphaned:\n{}", staleJobs);
+        }
 			}
+		} else {
+    	LOG.info("Not shutting down externally managed job executor.");
```
kagkarlsson: Indent is a bit messed up. I will probably do a reformat to 4 spaces after this PR is merged.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerBuilder.java
```diff
@@ -108,6 +104,11 @@ public SchedulerBuilder executorService(ExecutorService executorService) {
         return this;
     }
 
+    public SchedulerBuilder executorServiceShutdownTimeout(Duration shutdownTimeout) {
```
kagkarlsson: Not sure about the name. Especially since it is not applicable for the builder-method `executorService` where you send in the externally managed one.

What about `shutdownTimeout`?

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
```diff
@@ -26,348 +26,400 @@
...
+    return new SchedulerBuilder(dataSource, knownTasks);
+  }
+
+  private static ThreadPoolExecutor newThreadPool(String prefix, int maxThreads, Duration threadTimeout) {
+    ThreadPoolExecutor executor = new ThreadPoolExecutor(maxThreads, maxThreads,
+        threadTimeout.toMillis(), TimeUnit.MILLISECONDS,
+        new LinkedBlockingQueue<>(),
+        defaultThreadFactoryWithPrefix(prefix)
+    );
+    executor.allowCoreThreadTimeOut(true);
```
kagkarlsson: Currently, I don't think that the pool will ever be scaled to less than 3 threads? (Since currently all util-threads run until shutdown with their own sleep)

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
```diff
@@ -26,348 +26,400 @@
...
+    this.detectDeadWaiter = new Waiter(heartbeatInterval.multipliedBy(2), clock);
+    this.heartbeatInterval = heartbeatInterval;
+    this.heartbeatWaiter = new Waiter(heartbeatInterval, clock);
+    this.statsRegistry = statsRegistry;
+    this.pollingLimit = pollingLimit;
+
+    ThreadPoolExecutor housekeepingExecutor = newThreadPool(THREAD_PREFIX + "-housekeeper-", 3, THREAD_POOL_TIMEOUT);
+    this.dueExecutor = housekeepingExecutor;
+    this.detectDeadExecutor = housekeepingExecutor;
+    this.updateHeartbeatExecutor = housekeepingExecutor;
```
kagkarlsson: Not sure about this. The util-threads currently run "forever" (internal sleep). The only difference now will be that the thread-names will be a bit less describing. What do you feel is the benefit here?

Maybe they should be made into a `ScheduledExecutorService`, but I suspect that generates a bigger change..

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/ExecutorUtils.java
```diff
@@ -48,23 +48,34 @@ private static boolean awaitTermination(ExecutorService executor, Duration timeo
 	}
 
 	public static ThreadFactory defaultThreadFactoryWithPrefix(String prefix) {
-		return new PrefixingDefaultThreadFactory(prefix);
+		return new PrefixingDefaultThreadFactory(prefix, false);
+	}
+
+	public static ThreadFactory defaultThreadFactoryWithPrefix(String prefix, boolean useDaemonThreads) {
+		return new PrefixingDefaultThreadFactory(prefix, useDaemonThreads);
 	}
```
kagkarlsson: I can't seem to find any references to this method (where daemon=true)? Was there meant to be?

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
kagkarlsson: (the placement of this comment was probably better before I disabled whitespace changes... )

kagkarlsson:
Much of these changes have reached master via other PRs. Closing.

PR #99, opened 2020-04-27
allow rescheduling task with new data

pfrybar:
I have a use case where I need to reschedule a task, but with different data. `TaskRepository` already has this capability, so this PR just introduces a new method on the `Scheduler` and `SchedulerClient`.

Please let me know if this makes sense, or if you want me to change anything. Thanks!

### DIFF
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
@@ -148,6 +148,11 @@ public class Scheduler implements SchedulerClient {
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


kagkarlsson:
Makes total sense 👍 

Could you try and add a test for it to for example `SchedulerClientTest` ?

pfrybar:
Added a test that checks `reschedule(...)` with and without `newData`.

pfrybar:
That would be great! Thanks.

kagkarlsson:
`v6.9` on its way to central..

kagkarlsson:
Wil. update docs later

paul-frybarger:
👍 thanks for the quick review and merge

PR #103, opened 2020-05-08
Disable schedule startup tasks configuration option

balent:
In my use case, I'm creating recurring task dynamically from my application. I don't know which tasks should be running and what task data they should have during the startup.

With current behavior, all recurring tasks are started automatically with default task data, which is not desirable for me.

This option will disable recurring tasks to start on startup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,6 +216,8 @@ db-scheduler.scheduler-name=
 db-scheduler.threads=10
 # Ignored if a custom DbSchedulerStarter bean is defined
 db-scheduler.delay-startup-until-context-ready=false
+# Whether to schedule default recurring and startup tasks when application starts.
+db-scheduler.schedule-startup-tasks-enabled=true
 ```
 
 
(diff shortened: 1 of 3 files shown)


kagkarlsson:
Thanks for the contribution! 
Question tough, have you considered using a `CustomTask`? Would that solve your issue?

balent:
I checked out `CustomTask` but I couldn't figure out how to use it with cron schedule expression.

Instead of this configuration option, we can for example expose another function in RecurringTaskBuilder and allow to overwrite onStartup function. Something like this:

https://github.com/kagkarlsson/db-scheduler/pull/106

Then I can call execute with null onStartupHandler.

kagkarlsson:
Btw, using a custom task would look something like this (no onStartup is added via builder)

```
        Schedule recurringSchedule = FixedDelay.of(Duration.ofSeconds(4));
        CustomTask<Void> recurringNotStartup = Tasks.custom("recurring_no_startup", Void.class)
            .onFailureReschedule(recurringSchedule)  // default

                System.out.println("Executing " + taskInstance.getTaskAndInstance() + " , data: " + taskInstance.getData());
                    executionOperations.reschedule(executionComplete, recurringSchedule.getNextExecutionTime(executionComplete));
                };
            });
```

kagkarlsson:
Or better:

```
        Schedule recurringSchedule = FixedDelay.of(Duration.ofSeconds(4));
        CustomTask<Void> recurringNotStartup = Tasks.custom("recurring_no_startup", Void.class)
            .onFailureReschedule(recurringSchedule)  // default

                System.out.println("Executing " + taskInstance.getTaskAndInstance() + " , data: " + taskInstance.getData());
                return new CompletionHandler.OnCompleteReschedule<>(recurringSchedule);
            });
```

balent:
Thanks! That's what I was looking for.

PR #106, opened 2020-05-12
Allow RecurringTaskBuilder to set onStartup handler for each task

balent:


kagkarlsson:
I think a better variant here would be to add a builder-method to `RecurringTaskBuilder`, say
```
public RecurringTaskBuilder<T> skipScheduleAtStartup() {
...
}        
```

kagkarlsson:
Btw, using a custom task would look something like this (no onStartup is added via builder)

```
        Schedule recurringSchedule = FixedDelay.of(Duration.ofSeconds(4));
        CustomTask<Void> recurringNotStartup = Tasks.custom("recurring_no_startup", Void.class)
            .onFailureReschedule(recurringSchedule)  // default

                System.out.println("Executing " + taskInstance.getTaskAndInstance() + " , data: " + taskInstance.getData());
                    executionOperations.reschedule(executionComplete, recurringSchedule.getNextExecutionTime(executionComplete));
                };
            });
```

balent:
Thank you very much!

PR #111, opened 2020-05-27
Add TransferWise to list of users

BukhariH:
Hey Gustav + all!

Thanks so much for writing this library!

We've been using it for a good year in production at TransferWise & it's use has quickly expanded to a handful of services and a handful of different use cases and this library has never let us down or gotten in the way!


kagkarlsson:
That is fantastic to hear, really appreciate the feedback! Thanks!

PR #125, opened 2020-09-05
Update README.md polling interval default value

attilapapai:
Update default polling interval value from 10s to 30s in README


kagkarlsson:
Hi! Thank you for contributing!

I see that there is a mismatch between Spring Boot auto-starter and the default in [SchedulerBuilder](https://github.com/kagkarlsson/db-scheduler/blob/master/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerBuilder.java#L48). 

I am thinking that the Spring Boot default should be lowered to `10s` instead.. 🤔

PR #129, opened 2020-09-15
Spring Boot: Adjust defaults to align with main library

evenh:
Added public constants to `SchedulerBuilder`. When utilizing those constants the defaults in the starter will always be in sync the db-scheduler library. I've also bumped of Spring Boot to verify that we're good with the latest version.


kagkarlsson (approved):
Most excellent! 😄

Issue #124, opened 2020-09-01
Is it possible to change the cron schedule for existing task?


meghnaghai:
I have a query on updating schedule for the recurring task. Is it possible to change the cron schedule for existing task?

I created a task with a cron expression say every midnight. It ran for few days. Then I updated the cron expression to run the task every 5 min. But when I restart the application, it does not pickup the new schedule. I checked the db, and the next execution was scheduled for midnight only.

I was hoping the db scheduler should handle it. But I had to manually delete the record and restart application again in order to pick up the new schedule. Please let me know if I am missing something.

kagkarlsson:
Yes! If you change the `Schedule` and restart your app the execution in `scheduled_tasks` should have been updated with an execution-time matching the new `Schedule`

kagkarlsson:
What version are you using? In old versions, the execution-time was not updated until next time the task ran

meghnaghai:
I am using 7.2.

kagkarlsson:
How are you starting the scheduler and what does your task look like?

Important to use `startTasks(...)` for recurring tasks to get update execution-time on restart of `Scheduler`

```
final Scheduler scheduler = Scheduler
        .create(dataSource)
        .startTasks(hourlyTask)
        .build();
```

meghnaghai:
I am defining a bean for RecurringTask as below:
```
Tasks
      .recurring(
        TASK_NAME,
        new CronSchedule(cronExpression, ZoneId.of("UTC"))
      )
      .onFailureReschedule()
      .onDeadExecutionRevive()
      .execute((inst, ctx) -> {
        log.info("Processing Task in batch size of {}", batchSize);
        //task
      });
```

Havn't defined any specific config for scheduler.

kagkarlsson:
Double-checked and I remember incorrectly. You currently have to wait until the next execution-time. When that runs, the next selected execution-time will be according to the new schedule

kagkarlsson:
I think it should be possible to add a feature where the `Scheduler` checks existing records' execution-time and compares against the next execution-time according to the schedule. If the schedule indicates an earlier execution-time, update existing record.

At least for deterministic schedules such as `Daily` and `CronSchedule`. Unless the execution is currently running, update to next execution-time, and ideally next execution-time given the last time it ran.

meghnaghai:
Thanks for the clarification. Please let me know once the enhancement is released.

kagkarlsson:
Released fix in `v8.0`.

PR #136, opened 2020-09-23
Update existing execution for recurring task if schedule changes to more frequent

kagkarlsson:
If a `Schedule` for a `RecurringTask` is updated, update the future execution with the new execution-time to avoid having to wait for the execution to run to have execution-times based on new `Schedule`.

* For _deterministic_  schedule (Cron, Daily), update future execution for all changes
* For _non-deterministic_  schedule (FixedDelay), update future execution only if new execution-time is closer to `now()`


kagkarlsson:
Fixes #124

PR #143, opened 2020-10-12
Fix Spring Boot 2.2.0 deprecation (2.4.0 is coming)

evenh:
Since we already depend on Spring Boot 2.2+ (and using 2.3.x), I've took the liberty to modifiy the autoconfiguration to not depend on deprecated classes. Spring Boot 2.4 is scheduled for November 2020 and will remove deprecations from 2.2.0 (including `HealthIndicatorAutoConfiguration`).

See https://github.com/spring-projects/spring-boot/issues/22034


evenh:
See also this issue I've come across testing with milestone releases (M3+): https://github.com/spring-projects/spring-boot/issues/22988#issuecomment-707293527. Will hopefully be fixed before a stable release lands.

PR #149, opened 2020-10-26
Java 15 compatibility

evenh:
This PR will ensure that all tests will checked against AdoptOpenJDK 15 as well. Bumped `equalsverifier`, since it wasn't compatible with Java 15.


kagkarlsson (approved):
Nice, thanks! 🙇‍♂️

PR #153, opened 2020-11-13
Don't run AutodetectJdbcCustomization unless needed

matt-snider:
I noticed while debugging something else that the [`AutodetectJdbcCustomization constructor`](https://github.com/kagkarlsson/db-scheduler/blob/89626e6d1942d5e8ba6df4cdb758c72c50f0aa27/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/AutodetectJdbcCustomization.java#L34) is initialized by the builder every time, even if `JdbcCustomization` is already set. Since the constructor performs interactions with the DB it is better to put this in `orElseGet()` to prevent it unless needed.


kagkarlsson:
Ah, nice catch, thanks 👍

kagkarlsson (approved):
Merging 👍

PR #155, opened 2020-11-14
Scheduler client improvements

kagkarlsson:
* Change `getScheduledExecutions(..)` to return `List<..>` instead of taking a `Consumer`.
* The `Consumer` variant is renamed to `fetchScheduledExecution`.
* Introduce a `ScheduledExecutionsFilter` for `getScheduledExecutions(..)` to allow specifying whether or not to return picked executions

### DIFF
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/JdbcTaskRepository.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/JdbcTaskRepository.java
@@ -19,8 +19,10 @@ import com.github.kagkarlsson.jdbc.JdbcRunner;
 import com.github.kagkarlsson.jdbc.ResultSetMapper;
 import com.github.kagkarlsson.jdbc.SQLRuntimeException;
 import com.github.kagkarlsson.scheduler.TaskResolver.UnresolvedTask;
+import com.github.kagkarlsson.scheduler.jdbc.AndCondition;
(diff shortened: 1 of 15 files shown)


db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/JdbcTaskRepository.java
```diff
@@ -108,26 +108,37 @@ public boolean createIfNotExists(Execution execution) {
...
-    public void getScheduledExecutions(Consumer<Execution> consumer) {
+    public void getScheduledExecutions(ScheduledExecutionsFilter filter, Consumer<Execution> consumer) {
         UnresolvedFilter unresolvedFilter = new UnresolvedFilter(taskResolver.getUnresolved());
         jdbcRunner.query(
-                "select * from " + tableName + " where picked = ? " + unresolvedFilter.andCondition() + " order by execution_time asc",
+                "select * from " + tableName
+                    + (filter.getIncludePicked() && unresolvedFilter.unresolved.isEmpty() ? "" : " where ")
+                    + (!filter.getIncludePicked() ? " picked = ? " : "")
+                    + unresolvedFilter.andCondition()
+                    + " order by execution_time asc",
```
kagkarlsson, line 118: Need to refactor this, probably a bug in here..   (filter.getIncludePicked() + unresolvedFilter.andCondition())

cbelizone2y (approved):
Looks good to me

Issue #164, opened 2021-01-12
Task polling using sequential scan instead of index scan


rafal-kowalski:
Hi, we are using db-scheduler under quite heavy load (millions of due tasks, executing 2.5k tasks per second) and overall this library is handling it pretty well :) but sometimes throughput drops significantly (from 2.5k tsp to ~100 tsp). After our investigation we came to conclusion that the cause lies in Postgres query planner.
Since `TaskRepository.getDue` executes following sql code
`"select * from " + tableName + " where picked = ? and execution_time <= ? " + unresolvedFilter.andCondition() + " order by execution_time asc"`
Postgres query planner does not know that we need only first N tasks matching this criteria it decides to use sequential scan and begins sorting whole table based on execution_time as it would be faster in case when we need to get all of the rows, but when only subset is needed it could use presorted index, sometimes it does use index scan (query planner works in mysterious ways 😄).
Query execution times in our case look something like that:
- when sequential scan is used getDue takes around 10 seconds
- when index scan is used getDue takes around 50 milliseconds

Based on https://www.postgresql.org/docs/current/indexes-ordering.html when this kind of query has explicit LIMIT statement added it would always use index scan, as then it would know that we need only small subset.
With `JDBC.setMaxRows` limiting works in different way, in case of PostgreSQL driver it does not append LIMIT to query but after fetching desired row number it stops getting rest of data.
Since your library supports all SQL servers it might be impossible to implement generic way to always enforce this behaviour, afaik LIMIT won't work eg. in Oracle DB.

What we could do for our case is to override `JdbcTaskRepository.getDue()` and implement this method to work for Postgres.
I would love to hear another solutions or suggestions. 😃

kagkarlsson:
If it works as you describe I consider it a bug, and we need to find a way to solve it. (and your suggestion sounds good)

Though I am a bit surprised I haven't seen this before. I have run a load test against postgres and similarly had to create index etc, but in the end it achieved 2-5k executions/s. This was sometime ago, and I discovered some other bugs (race conditions) that could lead to the behavior you are describing (sudden drop in thoughput). Are you using the latest version?

Would you post your index here please?

(I also ran load tests using a new (unreleased) select-for-update strategy for postgres, which scales better, achieving 10k executions/s)

kagkarlsson:
Should we need to tune the sql, I would consider adding a config-option `.sqlUseExplicitLimit(true)` or something like that, and customize default JdbcTaskRepository accordingly

[...]

PR #165, opened 2021-01-15
Use explicit LIMIT in scheduler task polling query for PostgreSQL

rafal-kowalski:
Fixes behaviour described in https://github.com/kagkarlsson/db-scheduler/issues/164

### DIFF
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/JdbcTaskRepository.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/JdbcTaskRepository.java
@@ -135,9 +135,9 @@ public class JdbcTaskRepository implements TaskRepository {
     @Override
     public List<Execution> getDue(Instant now, int limit) {
         final UnresolvedFilter unresolvedFilter = new UnresolvedFilter(taskResolver.getUnresolved());
-
+        final String explicitLimit = jdbcCustomization.supportsExplicitQueryLimitPart() ? jdbcCustomization.getQueryLimitPart(limit) : "";
         return jdbcRunner.query(
-                "select * from " + tableName + " where picked = ? and execution_time <= ? " + unresolvedFilter.andCondition() + " order by execution_time asc",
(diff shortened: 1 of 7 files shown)


rafal-kowalski:
Ready for code review

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/JdbcTaskRepository.java
```diff
@@ -135,9 +135,9 @@ public void getScheduledExecutions(ScheduledExecutionsFilter filter, String task
     @Override
     public List<Execution> getDue(Instant now, int limit) {
         final UnresolvedFilter unresolvedFilter = new UnresolvedFilter(taskResolver.getUnresolved());
-
+        final String explicitLimit = jdbcCustomization.supportsExplicitQueryLimitPart() ? jdbcCustomization.getQueryLimitPart(limit) : "";
         return jdbcRunner.query(
-                "select * from " + tableName + " where picked = ? and execution_time <= ? " + unresolvedFilter.andCondition() + " order by execution_time asc",
+                "select * from " + tableName + " where picked = ? and execution_time <= ? " + unresolvedFilter.andCondition() + " order by execution_time asc" + explicitLimit,
```
kagkarlsson, line 140: I would be tempted to drop `setMaxRows(...)` when explicit limit in sql is used

kagkarlsson (commented):
Looks good 👍  
Just one small comment

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/JdbcTaskRepository.java
rafal-kowalski: Ok, I've removed `setMaxRows` when there is explicit limit

kagkarlsson (approved):
Looks good 👍

kagkarlsson:
Thank you for contributing!

PR #168, opened 2021-01-28
Add module db-scheduler-mongo

piemjean:
Hi,

I developped a specific module to implement a TaskRepository for MongoDB.
It relies on atomic operations, so that concurrency is handled. An index also ensures that tasks are unique.

I tried to keep things as separated as possible from the existing codebase. The only code alteration is in the parent pom.xml.

Thank you for your comments and for your great work.

### DIFF
--- a/db-scheduler-mongo/src/main/java/com/github/kagkarlsson/scheduler/MongoTaskRepository.java
+++ b/db-scheduler-mongo/src/main/java/com/github/kagkarlsson/scheduler/MongoTaskRepository.java
@@ -1,0 +1,434 @@
+package com.github.kagkarlsson.scheduler;
+
+import static com.github.kagkarlsson.scheduler.TaskEntity.Fields;
+import static com.mongodb.client.model.Filters.and;
+import static com.mongodb.client.model.Filters.eq;
+import static com.mongodb.client.model.Filters.lt;
+import static com.mongodb.client.model.Filters.lte;
+import static com.mongodb.client.model.Filters.ne;
+import static com.mongodb.client.model.Filters.nin;
+import static com.mongodb.client.model.Filters.or;
+import static com.mongodb.client.model.Sorts.ascending;
+import static com.mongodb.client.model.Updates.combine;
+import static com.mongodb.client.model.Updates.inc;
+import static com.mongodb.client.model.Updates.set;
+import static org.bson.codecs.configuration.CodecRegistries.fromProviders;
+import static org.bson.codecs.configuration.CodecRegistries.fromRegistries;
+
+import com.github.kagkarlsson.scheduler.TaskResolver.UnresolvedTask;
+import com.github.kagkarlsson.scheduler.task.Execution;
+import com.github.kagkarlsson.scheduler.task.Task;
+import com.github.kagkarlsson.scheduler.task.TaskInstance;
+import com.mongodb.ErrorCategory;
+import com.mongodb.MongoClient;
+import com.mongodb.MongoClientSettings;
+import com.mongodb.MongoWriteException;
+import com.mongodb.client.FindIterable;
+import com.mongodb.client.MongoCollection;
+import com.mongodb.client.MongoDatabase;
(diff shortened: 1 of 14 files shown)


kagkarlsson:
Hi! That is impressive and comprehensive work! I know mongodb-support has been requested by others as well.
I will need to think a bit about where this code should live. There is the slight problem that I may not be able to maintain MongoDb support very well myself, as I have no prior experience with Mongo 🤔. 
I can also see that it would beneficial with a better extension point for custom task-repositories so that you could avoid duplicating `Scheduler`, `SchedulerBuilder` etc

db-scheduler-mongo/src/main/java/com/github/kagkarlsson/scheduler/MongoTaskRepository.java
```diff
@@ -0,0 +1,434 @@
...
+
+        Bson update = combine(set(Fields.picked, true),
+            set(Fields.pickedBy, StringUtils.truncate(schedulerSchedulerName.getName(), 50)),
+            set(Fields.lastHeartbeat, timePicked),
+            inc(Fields.version, 1));
+
+        TaskEntity document = this.collection.findOneAndUpdate(filters, update, options);
+
+        return toExecution(document);
+    }
```
kagkarlsson, line 297: Does this method guarantee that for concurrent pick-operations, only one update will succeed? As far as I could tell it does, since it seem to have the version-filter and increments version on update. How does Mongo handle concurrent updates for a "row"? Only one suceeds?

piemjean: I have tested it, and Mongo can indeed make sure that only one thread can find the "row" that is (picked = false, version = n) and update it to (picked = true, version = n + 1), so that other threads won't fall in the same condition.
It is also describe in the example https://docs.mongodb.com/manual/tutorial/model-data-for-atomic-operations/.

kagkarlsson:
How would you feel if we moved this code to a separate github-project, for example one maintained by you? We can link to the project from the db-scheduler readme, from a list of community-maintained task-repositories.

If we keep it in this project repository I worry that as db-scheduler evolve, I may break things in the mongodb task-repository due unfamiliarity.

Additionally, I will put a backlog item for splitting out a clearer and better extension point for custom task-repositories. For example, a method on the builder where you can supply your own task-repository. And a better `TaskRepository` interface, declaring the contract for custom task-repositories.

piemjean:
This seems reasonnable since having a separate project is almost already in place. Instead of being a module, db-scheduler-mongo would pull db-scheduler as a dependency.
In terms of versionning, I think that following the main db-scheduler project is more simple, starting with the latest version: 9.3.

That would also be a first for me, so it might take a little while to completely set up the project on my end ;) 

I will update this PR once the project is properly initialized.

kagkarlsson:
Let me know if you need advice on setting up the project! 

Here's a guide on setting up deployment of artifacts to sonatype: https://central.sonatype.org/pages/ossrh-guide.html

kagkarlsson:
Created issue #173 for better support for custom task repositories

piemjean:
Thank you for the link to sonatype release guide, that was most helpful !

I released the first version with mongo at https://github.com/piemjean/db-scheduler-mongo/releases/tag/9.3

kagkarlsson:
Great job! I will close this PR once I have added your task-repository implementation as a link from the readme

Tetragramato:
Hi ! :) Is there any news on this feature ?

kagkarlsson:
I think this is working now, the code lives here: https://github.com/piemjean/db-scheduler-mongo

kagkarlsson:
Added a link to it in the readme:
https://github.com/kagkarlsson/db-scheduler/blob/master/README.md#third-party-task-repositories

Closing this PR as the code now lives in another github repo, `piemjean/db-scheduler-mongo`

kagkarlsson:

Issue #170, opened 2021-02-01
Failure logging


dmoidl:
Hi @kagkarlsson 👋 

I'm currently looking for a scheduler lib to use within our project and so I decided to give `db-scheduler` a try. So far it looks good and seems to cover pretty much everything we need at the moment. 

There is, however, one thing that bothers me and that could be a possible blocker for us: it logs each task failure as an `ERROR` ([here](https://github.com/kagkarlsson/db-scheduler/blob/9.3/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java#L359)). The way I see it, that doesn't make sense at all. I'd argue that from the perspective of the `Scheduler`, a task failure can hardly be an _error_, given the library itself provides a recovery mechanism (`FailureHandler`) used by the `Scheduler`.

I understand this may sound as a rather minor issue, but here's a real-life scenario to support my claim that it's not 🙂 : we plan to use the `db-scheduler` mainly for sending webhook notifications to registered third parties, each notification being a standard HTTP request to some URL. This means that such a task, i.e. a request, can easily fail and that's totally out of our control. Therefore we need to anticipate a failure and react accordingly - in our case by trying again later or disabling the webhook altogether if it failed for more than X times in a row.
Now, `db-scheduler` provides all the necessary tooling to achieve this, such as: 
 - the `CustomTask` type
 - the `FailureHandler` letting me reschedule the task or stop the execution based on the number of consecutive failures
 - the `Schedule` allowing me to easily implement type-task-agnostic exponential backoff mechanism for retries (nice 👍 )

However, if I do actually implement things in the way I described above, I simply _have to_ accept that on each task failure there will be an `ERROR` message in our logs. Taking into consideration that there may be rather high number of webhook notifications being sent each hour, day, etc. _and_ the failure likelihood of each is somewhat high, it would imply that our logs would be totally spammed by `ERROR`s produced by the `Scheduler`. This is obviously not acceptable.

I know I could work around this issue. I could disable logging for the `Scheduler` class for good, but then I wouldn't get even the real errors caused e.g. by misconfiguration. I could also _not_ use `FailureHandler` at all and basically duplicate the error-catching and handling mechanism directly in the `CompletionHandler`, but that kinda defeats the whole purpose of having the `FailureHandler` available 🙂 

[...]

PR #171, opened 2021-02-01
Change log level of error messages logged on execution failure from ERROR to DEBUG

dmoidl:
A solution for [this issue](https://github.com/kagkarlsson/db-scheduler/issues/170).


db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
```diff
@@ -356,12 +356,12 @@ private void executePickedExecution(Execution execution) {
                 statsRegistry.register(StatsRegistry.ExecutionStatsEvent.COMPLETED);
 
             } catch (RuntimeException unhandledException) {
-                LOG.error("Unhandled exception during execution of task with name '{}'. Treating as failure.", task.get().getName(), unhandledException);
+                LOG.debug("Unhandled exception during execution of task with name '{}'. Treating as failure.", task.get().getName(), unhandledException);
```
dmoidl, line 359: I can imagine also changing the log level in the default `FailureHandler` implementations (i.e. `OnFailureRetryLater` and `OnFailureReschedule`) to `ERROR` to keep pretty much the same experience for simple tasks using those built-ins, but it's not directly related to the problem this PR aims to solve so I didn't.

dmoidl: We discussed this and decided not to go this way.

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/logging/LogHelperTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.EnumSource;
+import org.slf4j.LoggerFactory;
+import org.slf4j.event.Level;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+
+public class LogHelperTest {
```
dmoidl, line 16: I hope a test of this kind is enough. I didn't really want to refactor much of the existing test code...

dmoidl (commented):
Here's the implementation for configurable failure logging as I promised 😉

db-scheduler/pom.xml
```diff
@@ -22,7 +22,7 @@
         <micro-jdbc.version>0.3</micro-jdbc.version>
         <otj-pg-embedded.version>0.11.4</otj-pg-embedded.version>
         <postgresql.version>9.4.1207</postgresql.version>
-        <slf4j.version>1.7.7</slf4j.version>
+        <slf4j.version>1.7.30</slf4j.version>
```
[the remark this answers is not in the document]
kagkarlsson: Will that work for users specifying an older version in their pom?

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerBuilder.java
```diff
@@ -163,6 +167,15 @@ public SchedulerBuilder commitWhenAutocommitDisabled(boolean commitWhenAutocommi
         return this;
     }
 
+    public SchedulerBuilder failureLogging(Level logLevel, boolean logStackTrace) {
+        if(logLevel == null) {
+            throw new IllegalArgumentException("Log level must not be null");
```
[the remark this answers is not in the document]
kagkarlsson: Looks good the way it is 👍

db-scheduler/pom.xml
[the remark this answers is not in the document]
kagkarlsson: On second thought I think it is better to not have a dependency on slf4f in the Builder-interface. I think we should create our own enum :)

db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/config/DbSchedulerProperties.java
```diff
@@ -106,6 +107,17 @@
...
+     * <p>Which log level to use when logging task failures. Defaults to {@link Level#DEBUG}.</p>
+     */
+    @NotNull
+    private Level failureLogLevel = Level.DEBUG;
+
+    /**
+     * <p>Whether or not to log the {@link Throwable} that caused a task to fail.</p>
+     */
+    private boolean logStackTraceOnFailure = false;
+
```
kagkarlsson, line 120: These need to refer to the constants in the `SchedulerBuilder` to ensure we get the same default

db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/config/DbSchedulerProperties.java
```diff
@@ -20,6 +20,7 @@
 import java.time.Duration;
 import java.util.Optional;
 import javax.validation.constraints.NotNull;
+import org.slf4j.event.Level;
```
kagkarlsson, line 23: As I mentioned, I think we need to create our own enum. It's cleaner :)

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
```diff
@@ -60,9 +62,12 @@
...
     private final SettableSchedulerState schedulerState = new SettableSchedulerState();
     private int currentGenerationNumber = 1;
+    private final LogHelper.LogMethod failureLogMethod;
+    private final boolean logStackTrace;
 
     protected Scheduler(Clock clock, TaskRepository schedulerTaskRepository, TaskRepository clientTaskRepository, TaskResolver taskResolver, int threadpoolSize, ExecutorService executorService, SchedulerName schedulerName,
-                        Waiter executeDueWaiter, Duration heartbeatInterval, boolean enableImmediateExecution, StatsRegistry statsRegistry, int pollingLimit, Duration deleteUnresolvedAfter, Duration shutdownMaxWait, List<OnStartup> onStartup) {
+                        Waiter executeDueWaiter, Duration heartbeatInterval, boolean enableImmediateExecution, StatsRegistry statsRegistry, int pollingLimit, Duration deleteUnresolvedAfter, Duration shutdownMaxWait, List<OnStartup> onStartup,
+                        Level logLevel, boolean logStackTrace) {
         this.clock = clock;
```
kagkarlsson, line 71: At some point I need to refactor this constructor 😊

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
```diff
@@ -82,6 +87,8 @@ protected Scheduler(Clock clock, TaskRepository schedulerTaskRepository, TaskRep
         this.updateHeartbeatExecutor = Executors.newSingleThreadExecutor(defaultThreadFactoryWithPrefix(THREAD_PREFIX + "-update-heartbeat-"));
         SchedulerClientEventListener earlyExecutionListener = (enableImmediateExecution ? new TriggerCheckForDueExecutions(schedulerState, clock, executeDueWaiter) : SchedulerClientEventListener.NOOP);
         delegate = new StandardSchedulerClient(clientTaskRepository, earlyExecutionListener);
+        this.failureLogMethod = LogHelper.getLogMethod(LOG, logLevel);
+        this.logStackTrace = logStackTrace;
```
kagkarlsson, line 91: We might make a bit more pretty if we introduce a wrapper-object, something like

```
failureLogger = new FailureLogger(LOG, logLevel, logStackTrace);
```
...
```
class FailureLogger
  logFailure(String message, Throwable, Object .. messageContext)
  ```

File: db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerBuilder.java [hunk 9]
```diff
@@ -190,7 +203,7 @@ public Scheduler build() {
             tableName,
             schedulerName.getName());
         return new Scheduler(clock, schedulerTaskRepository, clientTaskRepository, taskResolver, executorThreads, candidateExecutorService,
-                schedulerName, waiter, heartbeatInterval, enableImmediateExecution, statsRegistry, pollingLimit,
-            deleteUnresolvedAfter, shutdownMaxWait, startTasks);
+            schedulerName, waiter, heartbeatInterval, enableImmediateExecution, statsRegistry, pollingLimit,
+            deleteUnresolvedAfter, shutdownMaxWait, startTasks, logLevel, logStackTrace);
```
kagkarlsson, line 207: I think we should let `startTasks` be the last parameter here

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/logging/LogHelper.java
```diff
@@ -0,0 +1,50 @@
...
+import org.slf4j.Logger;
+import org.slf4j.event.Level;
+
+public class LogHelper {
+
+    public interface LogMethod {
+        void log(String format, Object... arguments);
+    }
+
+    public static LogMethod getLogMethod(Logger logger, Level logLevel) {
```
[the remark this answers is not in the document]
kagkarlsson: Yeah, nice use of method references 👍 , but see my suggestion on a `FailureLogger` class. (might still get to use some of this code though)

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/logging/LogHelperTest.java
kagkarlsson: Yeah, that should be fine. No need to verify in larger tests I think.

kagkarlsson (commented):
Looks good, but a couple of comments :) 

Btw, heads-up, I have some changes coming up as well. Depending on timing here, it might generate some small merge-work for this branch

db-scheduler/pom.xml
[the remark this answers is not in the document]
dmoidl: Done, new enum created.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/logging/LogHelper.java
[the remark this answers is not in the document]
dmoidl: Implemented the `FailureLogger` class as per your suggestion 🙂

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
dmoidl: Implemented.

dmoidl:
Sorry for force-pushing, but I had some conflicts with `master` and I prefer rebasing to keep the history clean 🙂

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/FailureLogger.java
```diff
@@ -0,0 +1,71 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.kagkarlsson.scheduler;
+
+import com.github.kagkarlsson.scheduler.logging.LogLevel;
+import org.slf4j.Logger;
+
+class FailureLogger {
```
dmoidl, line 21: I'd like to keep this class in the `logging` package, but at the same time I wanted it to be package-private as this is (arguably) not something we want to expose to the users... Or we could move the `LogLevel` to the main package and remove the `logging` package altogether. WDYT? 🤔

kagkarlsson: It doesn't really bother me if it would be public, I haven't really been strict about that in other places. Also, it turned out quite generic, so `.logging` sounds good

kagkarlsson: It became something more like `ConfigurableLogger` rather than something having anything to do with Failure :)

db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/config/DbSchedulerProperties.java
```diff
@@ -98,6 +100,16 @@
...
+    /**
+     * <p>Which log level to use when logging task failures. Defaults to {@link LogLevel#DEBUG}.</p>
+     */
+    private LogLevel failureLogLevel = SchedulerBuilder.DEFAULT_FAILURE_LOG_LEVEL;
+
+    /**
+     * <p>Whether or not to log the {@link Throwable} that caused a task to fail.</p>
+     */
+    private boolean logStackTraceOnFailure = SchedulerBuilder.LOG_STACK_TRACE_ON_FAILURE;
+
```
kagkarlsson, line 112: I was thinking if maybe these two properties should have the same prefix, so you can see they belong together in the spring properties file. Like `failureLoggerLevel`, `failureLoggerLogStacktrace` ?

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/FailureLogger.java
dmoidl: `ConfigurableLogger` it is, then.

db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/config/DbSchedulerProperties.java
dmoidl: I was thinking the same but couldn't come up with a good name for the latter property 🙃  The ones that you proposed sound good though, code updated.

dmoidl:
Looks like the failing checks are not caused by my changes...

kagkarlsson:
Upgrading to testcontainers 1.15.1, that should fix it.

kagkarlsson:
Having some seconds thoughts on what should be the default log-level. Would not WARN be a safer default, and then users' who are bothered by it can tune it explicitly to DEBUG? I worry that some users' will miss important exceptions using the default-config (i.e. it is not "safe")

dmoidl:
> Having some seconds thoughts on what should be the default log-level. Would not WARN be a safer default, and then users' who are bothered by it can tune it explicitly to DEBUG? I worry that some users' will miss important exceptions using the default-config (i.e. it is not "safe")

Sure, I think this all boils down to how you'd like the people to use the library 🙂  In my eyes, it really doesn't matter all that much since now everyone can easily adjust the logging as they see fit, but you're right that `DEBUG` may be too low if you want to provide some information to the user using the out-of-the-box settings.

I've changed the default to `WARN` and also added a new log level `OFF` to turn the logging off completely so that it's 100% customizable 🙂

kagkarlsson:
Excellent! I will merge this very soon 👍

dmoidl:
@kagkarlsson Sorry, didn't notice those failing tests. Fix pushed 🙂

kagkarlsson:

PR #179, opened 2021-02-18
Add some toStrings to 

runeflobakk:
Small contribution. Basically adding some more informative toString implementations to:
- `CronSchedule`, because it was missing. Already present in Daily and FixedDelay.
- Task implementations: adds the _type_ of task (RecurringTask, OneTimeTask etc), and also try to output some information about rescheduling from RecurringTask.
- Keeping the toString in `Task` to have a base toString for custom derivations from Task. Small fix from `task=` to `name=` to align with what seemed to be the existing style/format.

As the toString implementation for RecurringTask may seem a little... assuming perhaps (?), I am open to suggestions on how to better structure it, if you see any problems with it.

A bit on the side, but I also wanted to bring [ThreeTen Extra](https://www.threeten.org/threeten-extra/) and its [AmountFormats](https://www.threeten.org/threeten-extra/apidocs/org.threeten.extra/org/threeten/extra/AmountFormats.html) to your attention for some more human-readable outputting of `Duration`. Small, focused and stable library. I assume it could be easily shaded as you already do with cronutils. You would need to include the `/org/threeten/extra/wordbased_en.properties` resource. I can open up a separate issue for it if you want, but wanted to mention it here in the context of more informative toStrings and logging.

### DIFF
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/helper/RecurringTask.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/helper/RecurringTask.java
@@ -57,4 +57,9 @@ public abstract class RecurringTask<T> extends Task<T> implements OnStartup {
 
     public abstract void executeRecurringly(TaskInstance<T> taskInstance, ExecutionContext executionContext);
 
+    @Override
+    public String toString() {
+        return "RecurringTask name=" + getName() + ", scheduled using " + onComplete;
+    }
+
 }
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/helper/RecurringTask.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/helper/RecurringTask.java
@@ -59,7 +59,7 @@ public abstract class RecurringTask<T> extends Task<T> implements OnStartup {
 
     @Override
     public String toString() {
-        return "RecurringTask name=" + getName() + ", scheduled using " + onComplete;
+        return "RecurringTask name=" + getName() + ", onComplete=" + onComplete;
     }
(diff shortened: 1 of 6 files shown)


kagkarlsson:
Excellent, thank you for contributing! :) 
Will need to think a bit on bringing in a shaded dependency for better to-stringing 🤔

kagkarlsson (approved):
💯

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/helper/RecurringTask.java
```diff
@@ -57,4 +57,9 @@ public void onStartup(Scheduler scheduler, Clock clock) {
 
     public abstract void executeRecurringly(TaskInstance<T> taskInstance, ExecutionContext executionContext);
 
+    @Override
+    public String toString() {
+        return "RecurringTask name=" + getName() + ", scheduled using " + onComplete;
```
kagkarlsson, line 62: Alternatively
```
"RecurringTask name=" + getName() + " with onComplete=" + onComplete;
```

runeflobakk: Yes, I agree. That may be, strictly speaking, more correct, as a task itself does not know anything about it actually being scheduled or not. Maybe even just
```suggestion
        return "RecurringTask name=" + getName() + ", onComplete=" + onComplete;
```
to align with the format of other toStrings, I don't know if you have any preference on that. I myself tend to like some small bits of "prose"/sentence-like toStrings, instead of enumerating comma-separated key/value pairs, as I think it make them easier to read in log messages, but that is an individual preference of mine.

PR #182, opened 2021-02-19
Spring Boot Devtools-friendly default task data serializer

k-jamroz:
Default java serializer does not work correctly with Spring Boot Devtools. During deserialization `ClassCastException`s are thrown. That is because `ObjectInputStream` does not use correct class loader (classloaders in devtools are more complicated). See: https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#using-boot-devtools-known-restart-limitations

### DIFF
--- a/db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/autoconfigure/DbSchedulerAutoConfiguration.java
+++ b/db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/autoconfigure/DbSchedulerAutoConfiguration.java
@@ -18,6 +18,7 @@ package com.github.kagkarlsson.scheduler.boot.autoconfigure;
 import com.github.kagkarlsson.scheduler.Scheduler;
 import com.github.kagkarlsson.scheduler.SchedulerBuilder;
 import com.github.kagkarlsson.scheduler.SchedulerName;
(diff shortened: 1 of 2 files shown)


db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/autoconfigure/DbSchedulerAutoConfiguration.java
```diff
@@ -203,4 +211,38 @@ private static DataSource configureDataSource(DataSource existingDataSource) {
...
+            } catch (Exception e) {
+                throw new RuntimeException("Failed to serialize object", e);
+            }
+        }
+
+        public <T> T deserialize(Class<T> clazz, byte[] serializedData) {
+            if (serializedData == null)
+                return null;
+            try (ByteArrayInputStream bis = new ByteArrayInputStream(serializedData);
+                 ObjectInput in = new ConfigurableObjectInputStream(bis, Thread.currentThread().getContextClassLoader())) {
```
k-jamroz, line 240: this is the only difference with `Serializer.DEFAULT_JAVA_SERIALIZER`

kagkarlsson:
Thank you for contributing! Will have a look as soon as I get the time. I have heard about this from other users, so I appreciate a fix 👍

kagkarlsson (approved):
LGTM 👍

PR #184, opened 2021-02-21
Adding the ability to trigger task run manually.

FreCap:
Motivation: in processing pipelines where tasks can be time-intensive, and multiple queues/tasks are fighting for resources, the lib users can benefit from implementing a more complex pulling strategy and trigger executions manually. Hence we start exposing a public executeDueTasks() method.

Changes:
- Exposing executeDueTasks as public
- renaming Scheduler.start() to Scheduler.startConsumer(), keeping backward compatibility, to reflect that only the "consumer-machines" will have to start this process
- add the ability to disable consuming with SchedulerBuilder.disableStartConsumingTasksOnStart(), in such a way that lifecycle threads can handle heartbeat and dead executions, but not computation heavy threads (of the executions), which can be triggered manually

Additional:
- feature tested
- all methods changes are backward compatible.

### DIFF
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/DueExecutionsBatch.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/DueExecutionsBatch.java
@@ -15,27 +15,39 @@
  */
 package com.github.kagkarlsson.scheduler;
 
+import java.util.concurrent.CompletableFuture;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.Supplier;
+
 
 class DueExecutionsBatch {
 
     private static final Logger LOG = LoggerFactory.getLogger(DueExecutionsBatch.class);
     private final int generationNumber;
     private final AtomicInteger executionsLeftInBatch;
+    private final CompletableFuture<BatchCompletionResult> onBatchComplete;
     private int threadpoolSize;
     private boolean possiblyMoreExecutionsInDb;
     private boolean stale = false;
     private boolean triggeredExecuteDue;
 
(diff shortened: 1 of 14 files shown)


kagkarlsson:
Thank you for contributing!

I think I need to understand this use-case better in order to evaluate this feature/pull-request. Could you describe some scenario where this would be useful (in more detail than above)? I assume you can't just increase the thread-count?

Also, a small heads-up here, this will likely have some conflicts with another PR I am looking to merge soon: https://github.com/kagkarlsson/db-scheduler/pull/175

FreCap:
@kagkarlsson thanks for the comment!

One of the use cases, is when you have a set of tasks that always have to be run before the others.
Let's say you have 3 different tasks:
1) "put off fire"
2) "send some texts"
3) "get other"

You want to run all of the "put off fire" fire before going to any other task. 
All of those tasks have different dataclasses.

Having `executeDueTasks ` can facilitate greatly situations where we have to handle this case by building a coordinator

kagkarlsson:
> You want to run all of the "put off fire" fire before going to any other task.
> All of those tasks have different dataclasses.

Wouldn't this be solved by adding the priority-feature?

Also, do you mean that 1) must run completely before 2) should start running?

FreCap:
In the other post you put it in better words than mine:
`Do you require to be able to set priority on instance-level, or would task-level work? I.e., for a given task, all instances have the same priority`, both are needed

The difference is that at instance-level once the priority is given, that is the priority.

At task level instead it can be flexible. In the sense that a given point in time the task "compute-news-midnight" with all of its instances has more priority than "compute-news-midday", and another point in time the other is more important to complete

kagkarlsson:
Check out ManualScheduler



fre. 15. okt. 2021 kl. 21:39 skrev maratik123 ***@***.***>:

PR #219, opened 2021-07-28
Add Loadster to the list of companies using db-scheduler.

azhawkes:
This scheduler is working great for us at Loadster and we'd like to be added to the list! Happily running in production for over a month now and a big improvement over our old task scheduler. Thanks @kagkarlsson.


kagkarlsson:
Great to hear!! 🙏

PR #225, opened 2021-09-04
Enhance ExecutePicked logs. Fixes kagkarlsson/db-scheduler#223

### LINKED ISSUE #223: Bad-looking ExecutePicked logging
`com.github.kagkarlsson.scheduler.ExecutePicked` [logs](https://github.com/kagkarlsson/db-scheduler/blob/master/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/ExecutePicked.java#L82) the start of an execution like this:
> Executing Execution: task=example-task-name, id= ...

As you can see, there are 2 issues:
* Duplicated **_Executi___**
* Redundant capital letter in **_Execution_**

I suggest changing it to one of two opitions:

1. >  Starting execution: task=example-task-name, id= ...

2. > Executing: task=example-task-name, id= ...

The change would require modifying the [toString()](https://github.com/kagkarlsson/db-scheduler/blob/master/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/Execution.java#L75) method of `com.github.kagkarlsson.scheduler.task.Execution` or altering what it returns before a log message is created.

I use the `ExecutePicked` logs frequently while debugging db-scheduler so it would be nice to make the log message look better.

adam-handke:
The debug logs of `ExecutePicked` for starting and finishing an execution are enhanced to be more readable and informative.

### DIFF
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/ExecutePicked.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/ExecutePicked.java
@@ -79,9 +79,12 @@ class ExecutePicked implements Runnable {
 
         Instant executionStarted = clock.now();
         try {
-            LOG.debug("Executing " + execution);
+            String executionString = execution.toString();
+            String decapitalizedExecutionString = Character.toLowerCase(executionString.charAt(0)) + executionString.substring(1);
+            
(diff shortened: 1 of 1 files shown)


kagkarlsson:
Fixes outside PR. Closing. Thank you for contributing!

Issue #228, opened 2021-09-22
Stopping the scheduler takes several seconds.


sargue:
This simple test says the stop took 3 seconds.

```
public class TestSchedulerShutdown {
    @Test
    public void test() {
        JdbcDataSource ds = new JdbcDataSource();
        ds.setURL("jdbc:h2:mem");
        ds.setUser("sa");
        ds.setPassword("sa");
        var scheduler = Scheduler.create(ds).build();
        scheduler.start();
        var t0 = System.currentTimeMillis();
        scheduler.stop();
        System.out.printf("Stop took %d ms%n", System.currentTimeMillis() - t0);
    }
}
```

A quick look at the code seems that the shutdown of the internal executors relies on some timeouts. Some wake up / signaling should make this process not relying on timeouts or polling.

kagkarlsson:
Yeah, agreed.

kagkarlsson:
Released in 10.4

sargue:
I've retried the same exact code just bumping the version to 10.5 and the delay, indeed, is lower but not zero.

I'm still getting 1 second delay to stop. It's quite consistent around 1000-1100 ms so it looks like some 1000 ms timeout is there still.

PR #231, opened 2021-10-09
Use ScheduledExecutorService for housekeeping tasks for faster shutdown.

kagkarlsson:
For faster shutdown, change from many single-threaded ExecutorServices to one `ScheduledExecutorService` for util background jobs. 

Also, shutdown `dueExecutor` interruptibly diredctly if it is currently sleeping. Again to achieve faster shutdown.


kagkarlsson:
Fixes #228

PR #234, opened 2021-10-14
Don't require Spring Boot Actuator to work

evenh:
I recently found out that the Spring Boot starter implicitly requires [Actuator](https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html) to work. This PR changes the autoconfiguration so it will work both with and without Actuator on the classpath.


kagkarlsson (approved):
Looks good to me 👍 🙏

PR #267, opened 2022-02-22
change transferwise to wise in the readme due to renaming

aleksandrbocharov:
Hey 
we have renamed from TransferWise to Wise some time ago 🙂 
so changing the readme to reflect  this


kagkarlsson:
Thanks for updating the Readme 👍 

I found [tw-tasks-executor](https://github.com/transferwise/tw-tasks-executor) some time ago, so was thinking that maybe you had migrated away from db-scheduler. Assuming this is not the case now :)

anikkonm (approved):

PR #289, opened 2022-05-31
CI: Remove deprecated GitHub Actions

evenh:
Reduces number of deprecation warnings. Also adds support for Java 17 (and possibly new versions in the future).

### DIFF
--- a/db-scheduler/pom.xml
+++ b/db-scheduler/pom.xml
@@ -18,7 +18,8 @@
         <hikaricp.version>4.0.3</hikaricp.version>
         <hsqldb.version>2.6.1</hsqldb.version>
         <java8-matchers.version>1.6</java8-matchers.version>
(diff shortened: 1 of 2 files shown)


db-scheduler/pom.xml
```diff
@@ -205,14 +212,34 @@
...
+        <dependency>
+            <groupId>net.bytebuddy</groupId>
+            <artifactId>byte-buddy</artifactId>
+            <version>${bytebuddy.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>net.bytebuddy</groupId>
+            <artifactId>byte-buddy-agent</artifactId>
+            <version>${bytebuddy.version}</version>
+        </dependency>
```
kagkarlsson, line 242: Should it be `scope=test` on these? 🤔

kagkarlsson (approved):
Looking good, thank you! :)

PR #301, opened 2022-07-05
Changing ExecutionComplete constructor visibility to public

Cesaario:
I don't know exactly if it was intended, but I didn't find a way to reschedule a task from its dead execution handler.
I tried creating a new handler and passing it to ```onDeadExecution```, but its parameters are ```Execution``` and ```ExecutionOperation<Void>```.
Reading the code from ```DeadExecutionHandler.java``` I noticed that it creates a new ```ExecutionComplete```, passing ```execution``` as a parameter. I expected to be able to do the same thing, but ```ExecutionComplete``` has a constructor with default visibility (package only).
image

I don't know if ```ExecutionComplete```'s constructor was supposed to be only visibile from its package. If not, this PR should fix it. If the behavior is intented, how should I reschedule my task from its dead execution handler?

### DIFF
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/ExecutionComplete.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/ExecutionComplete.java
@@ -26,7 +26,7 @@ public class ExecutionComplete {
     private final Result result;
     private final Throwable cause;
 
-    ExecutionComplete(Execution execution, Instant timeStarted, Instant timeDone, Result result, Throwable cause) {
+    public ExecutionComplete(Execution execution, Instant timeStarted, Instant timeDone, Result result, Throwable cause) {
         this.timeStarted = timeStarted;
         this.cause = cause;
         if (result == Result.OK && cause != null) {


kagkarlsson:
Yeah I think that is a bug in the api. I'd rather the `ExecutionComplete` came as a parameter to `deadExecution`, but that is a breaking change I have to consider some.. For now I think we can make the constructor public as you suggest

kagkarlsson:
What is the exact operation you are trying to achieve in the dead execution handler?

Cesaario:
At first I wanted to increase the consecutive_failures counter when a dead execution was detected, but then I realized that the API already does that. But still, I think it would be nice to be able to perform some actions and still be able to reschedule the task in the dead execution handler, similar to what is possible with the onFailure handler:
image

kagkarlsson:
Created related issue #302

Cesaario:
> Looking at the code again, there is a public static factory-method for it:
> 
> https://github.com/kagkarlsson/db-scheduler/blob/master/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/ExecutionComplete.java#L44
> 
> Can you use that?

Oh, I didn't notice that method. Yes, I can use that, it worked perfectly. Thanks!

kagkarlsson:
I suppose this can be closed?

Cesaario:
Sure!

Issue #114, opened 2020-06-04
Async, non-blocking jobs execution


coutoPL:
Hi,

Your lib almost perfectly fits my use case. But, after code inspecting, I suppose there is no support for asynchronous, non blocking jobs. 

What do I exactly mean? I have plenty of jobs to start (let's say 10k/min). Each of it contains a HTTP request(s), which I call using non-blocking http client. So, I have eg. `CompletableFuture[Response]` and based on the response I have to decide if reschedule the task instance or not (custom task and `ExecutionHandler` seem to be great to do it). Currently, I have to block the thread, but this is not the way to go, because there is a lot of tasks to start (requirement: immediately or as soon as possible) and mean time of waiting for response can be ~30s. The thread is blocked and just waiting for IO. 

It seems that it can be solved if db-scheduler could be able to define and use sth like async execution handler:

```
public interface ExecutionHandler<T> {
   CompletableFuture<CompletionHandler<T>> execute(TaskInstance<T> taskInstance, ExecutionContext executionContext);
}
```

Or maybe I missed sth and there is a way to achieve sth like described above?
If not, WDYT about the idea to introduce async interface and adapt db-scheduler to be able to work with it? Do you see any obstacles?

Thanks for your effort. This project and your care for it looks very impressive.

kagkarlsson:
Hi! What level of parallelism per scheduler-instance do you need? 

I haven't really thought about it before, but I agree it makes sense if you have many but slow executions. Trying to think about what the obstacles are...

kagkarlsson:
I don't see any obvious obstables, probably need to get a feel for how it would affect the code-base to be able to evaluate it properly

coutoPL:
I'd like to control parallelism by thread pool set for a scheduler instance. 

Let's say I give the scheduler pool of 5 threads (fixed) and the scheduler should run as many executions as it can. If job/execution release thread (because eg. it waits for IO), the next job/execution can be run. The IO result of previous execution will be (can be) handled (after async response) by other thread than the initiation one.

Ok, if we decide to use db-scheduler, I'll try to add this functionality to the lib, if you don't mind :)

kagkarlsson:
I am a bit uncertain about how the algorithm controlling how many executions the scheduler would be allowed to pick would look... 🤔

coutoPL:
Me neither. I've not been there yet. But I'm going to consult with you all uncertain things here. Stay tuned :)

muradm:
[...]

Issue #116, opened 2020-06-25
Helpful stuck-thread logging


kagkarlsson:
Add some helpful logging for executions that run for a long time and might be "stuck". Include call-stack of thread in logging.

Config: 
* enable by default, allow disable
* configurable threshold for what is considered "a long time"

PR #304, opened 2022-07-07
Enhance db-scheduler to support asynchronous task execution

amit-handda:
Thank you for maintaining db-scheduler.

# What
This PR enhances db-scheduler to support task execution asynchronously. It addresses [this issue](https://github.com/kagkarlsson/db-scheduler/issues/114)

# Why
db-scheduler has an executorservice of threadpool. Threadpools are used to schedule db-scheduler tasks.
Tasks were executed in a synchronous way. If tasks do something (http call eg) asynchronously, the threadpool will remain blocked.
This might not be an issue where tasks are tiny or the task throughput is tiny. But, it can become an issue for some scenarios.

# Current status
* Few TCs are flaky due to asychronous implementation. To look into fixing it.
* Create PR upstream to the original repo, get it merged possibly

### DIFF
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/ExecutePicked.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/ExecutePicked.java
@@ -29,6 +29,7 @@ import org.slf4j.LoggerFactory;
 import java.time.Instant;
 import java.util.Optional;
 import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
 
 @SuppressWarnings("rawtypes")
 class ExecutePicked implements Runnable {
@@ -61,39 +62,37 @@ class ExecutePicked implements Runnable {
     public void run() {
         // FIXLATER: need to cleanup all the references back to scheduler fields
         final UUID executionId = executor.addCurrentlyProcessing(new CurrentlyExecuting(pickedExecution, clock));
-        try {
-            statsRegistry.register(StatsRegistry.CandidateStatsEvent.EXECUTED);
-            executePickedExecution(pickedExecution);
-        } finally {
-            executor.removeCurrentlyProcessing(executionId);
-        }
+        statsRegistry.register(StatsRegistry.CandidateStatsEvent.EXECUTED);
+        executePickedExecution(pickedExecution).whenComplete((c, ex) -> executor.removeCurrentlyProcessing(executionId));
     }
 
-    private void executePickedExecution(Execution execution) {
+    private CompletableFuture<CompletionHandler> executePickedExecution(Execution execution) {
         final Optional<Task> task = taskResolver.resolve(execution.taskInstance.getTaskName());
         if (!task.isPresent()) {
             LOG.error("Failed to find implementation for task with name '{}'. Should have been excluded in JdbcRepository.", execution.taskInstance.getTaskName());
             statsRegistry.register(StatsRegistry.SchedulerStatsEvent.UNEXPECTED_ERROR);
-            return;
+            return new CompletableFuture<>();
         }
 
         Instant executionStarted = clock.now();
-        try {
-            LOG.debug("Executing " + execution);
-            CompletionHandler completion = task.get().execute(execution.taskInstance, new ExecutionContext(schedulerState, execution, schedulerClient));
-            LOG.debug("Execution done");
+        LOG.debug("Executing " + execution);
+        CompletableFuture<CompletionHandler> completableFuture = task.get().execute(execution.taskInstance, new ExecutionContext(schedulerState, execution, schedulerClient));
 
+        return completableFuture.whenCompleteAsync((completion, ex) -> {
+            if (ex != null) {
+                if (ex instanceof RuntimeException) {
+                    failure(task.get(), execution, ex, executionStarted, "Unhandled exception");
+                    statsRegistry.register(StatsRegistry.ExecutionStatsEvent.FAILED);
+                } else {
+                    failure(task.get(), execution, ex, executionStarted, "Error");
+                    statsRegistry.register(StatsRegistry.ExecutionStatsEvent.FAILED);
+                }
+                return;
+            }
+            LOG.debug("Execution done");
             complete(completion, execution, executionStarted);
             statsRegistry.register(StatsRegistry.ExecutionStatsEvent.COMPLETED);
-
-        } catch (RuntimeException unhandledException) {
-            failure(task.get(), execution, unhandledException, executionStarted, "Unhandled exception");
-            statsRegistry.register(StatsRegistry.ExecutionStatsEvent.FAILED);
-
-        } catch (Throwable unhandledError) {
-            failure(task.get(), execution, unhandledError, executionStarted, "Error");
-            statsRegistry.register(StatsRegistry.ExecutionStatsEvent.FAILED);
-        }
+        }, executor.getExecutorService());
     }
 
     private void complete(CompletionHandler completion, Execution execution, Instant executionStarted) {
--- a/db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/SchedulerClientTest.java
+++ b/db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/SchedulerClientTest.java
@@ -16,6 +16,7 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
 import java.time.Instant;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import static java.time.Duration.ofSeconds;
@@ -65,16 +66,18 @@ public class SchedulerClientTest {
     }
 
     @Test
-    public void client_should_be_able_to_schedule_executions() {
+    public void client_should_be_able_to_schedule_executions() throws InterruptedException {
         SchedulerClient client = SchedulerClient.Builder.create(DB.getDataSource()).build();
         client.schedule(oneTimeTaskA.instance("1"), settableClock.now());
 
         scheduler.runAnyDueExecutions();
+        // Since execution is executed in an async way, we need to wait for a while to let the execution finish before asserting
+        Thread.sleep(1000);
         assertThat(onetimeTaskHandlerA.timesExecuted.get(), CoreMatchers.is(1));
     }
 
     @Test
-    public void should_be_able_to_schedule_other_executions_from_an_executionhandler() {
+    public void should_be_able_to_schedule_other_executions_from_an_executionhandler() throws InterruptedException {
         scheduler.schedule(scheduleAnotherTask.instance("1"), settableClock.now());
         scheduler.runAnyDueExecutions();
         assertThat(scheduleAnother.timesExecuted, CoreMatchers.is(1));
@@ -82,6 +85,8 @@ public class SchedulerClientTest {
 
         scheduler.tick(ofSeconds(1));
         scheduler.runAnyDueExecutions();
+        // Since execution is executed in an async way, we need to wait for a while to let the execution finish before asserting
+        Thread.sleep(1000);
         assertThat(onetimeTaskHandlerA.timesExecuted.get(), CoreMatchers.is(1));
     }
 
@@ -155,9 +160,10 @@ public class SchedulerClientTest {
         }
 
         @Override
-        public void execute(TaskInstance<T> taskInstance, ExecutionContext executionContext) {
+        public CompletableFuture<Void> execute(TaskInstance<T> taskInstance, ExecutionContext executionContext) {
             executionContext.getSchedulerClient().schedule(secondTask, instant);
             this.timesExecuted++;
+            return CompletableFuture.completedFuture(null);
         }
     }
 }
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Executor.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Executor.java
@@ -99,4 +99,8 @@ public class Executor {
             LOG.warn("Released execution was not found in collection of executions currently being processed. Should never happen. Execution-id: " + executionId);
         }
     }
+
+    public ExecutorService getExecutorService() {
+        return executorService;
+    }
 }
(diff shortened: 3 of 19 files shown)


kagkarlsson:
Thank you! I am currently on vacation and will not be able to look at it for a couple of weeks. And it is also a critical change so I need some time to consider it

x0a1b:
I see some significant performance improvements as well. @amit-handda can you post some benchmarks as well?

amit-handda:
As expected, we have observed signficant throughput gains for async tasks, as follows:

  | Current (execution/s) | Future-based (execution/s)
-- | -- | --
1 scheduler instance |   |  
300 threads, 10k tasks | 285 | 1,666
4 scheduler instances |   |  
20 threads, 10k tasks | 80 | 2,206
100 threads, 10k tasks | 359 | 2,407

Task def
image

amit-handda:
Hello @kagkarlsson, hope you had nice vacation. Wondering if you are back and had the time to look at the PR ? Let me know if you have any query. TY !

kagkarlsson:
Thank you! Yes I am back, but I struggle to get the bandwidth to review this PR unfortunately. 

I have scanned through the code and noted a couple of things: 

* I am not sure how I feel about breaking the  `ExecutionHandler` interface. If the async variant could possibly be introduced more gently, as an option rather.. 
* Are we allowing more things to happen in parallel? (If so, I am not sure what the consequences of this will be)

amit-handda:
Hello @kagkarlsson , thanks for the feedback. ques
1/ how do you intend the `ExecutionHandler` interface to be ?
The PR is allowing the `ExecutionHandler` implementation to return a Future<CompletionHandler> so that the db-scheduler thread would be unblocked (in case the implementation performs async ops such as HTTP calls etc.)
Thanks,

kagkarlsson:
> 1/ how do you intend the ExecutionHandler interface to be ?

Well, that is the problem, I do not know currently. What options do we have 🤔

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/ExecutePicked.java
```diff
@@ -61,39 +62,37 @@ public ExecutePicked(Executor executor, TaskRepository taskRepository, Scheduler
...
 
         Instant executionStarted = clock.now();
-        try {
-            LOG.debug("Executing " + execution);
-            CompletionHandler completion = task.get().execute(execution.taskInstance, new ExecutionContext(schedulerState, execution, schedulerClient));
-            LOG.debug("Execution done");
+        LOG.debug("Executing " + execution);
+        CompletableFuture<CompletionHandler> completableFuture = task.get().execute(execution.taskInstance, new ExecutionContext(schedulerState, execution, schedulerClient));
 
+        return completableFuture.whenCompleteAsync((completion, ex) -> {
```
kagkarlsson, line 81: Not 100% sure what happens here. What will the significance of `whenCompleteAsync` be here?

Will it happen in parallell with  `executor.removeCurrentlyProcessing` here? 
```
executePickedExecution(pickedExecution).whenComplete((c, ex) -> executor.removeCurrentlyProcessing(executionId));
```

kagkarlsson:
> As expected, we have observed signficant throughput gains for async tasks, as follows:
> 
>  	Current (execution/s)	Future-based (execution/s)
> 1 scheduler instance	 	 
> 300 threads, 10k tasks	285	1,666
> 4 scheduler instances	 	 
> 20 threads, 10k tasks	80	2,206
> 100 threads, 10k tasks	359	2,407


I need to understand why a future-based execution will be that much faster. Feels like we are skipping av step..

amit-handda:
> 
> I need to understand why a future-based execution will be that much faster. Feels like we are skipping av step..
Hi, It is due to our task defn. Task implementation does nothing other than sleeping for a second. After our changes, the sleeping tasks dont block the scheduler threadpool (since task implementation returns futures). Before our changes, same task implementation would block the scheduler threadpool (hence, leading to inferior performance), let me know if you have any issues. 
https://user-images.githubusercontent.com/56566242/182731346-cf142ba0-56f5-45ad-a888-69b8c4378aac.png

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/ExecutePicked.java
amit-handda: using `whenCompleteAsync` to associate the `action` to the db-scheduler executor explicitly (otherwise, it might use the common forkjoinpool).
I dont think it ll execute in parallel to `removeCurrentlyProcessing` since `*Async` API only enables to specify the executor service for the action.
ref link: https://www.linkedin.com/pulse/asynchronous-programming-java-completablefuture-aliaksandr-liakh/
please let me know in case I am off.

amit-handda:
> Well, that is the problem, I do not know currently. What options do we have 🤔

IMO, returning a `completablefuture<completionhandler>` is the only option to enable task to execute async operations and unblocking the db-scheduler threadpool. <- this necessitates updating the interface in a non-gentle way. :)

maxpert:
I think what @amit-handda is mentioning is more specific for reactive solutions (In this case Kotlin coroutines which are non-blocking on threads). I think having this in place will unlock bunch of reactive usecase (including Spring framework users) to write their workers in more scalable way.

kagkarlsson:
The results you are referring to, where future-based and non-future based tasks are compared, what polling-strategy are they using, and what limits?

maxpert:
> The results you are referring to, where future-based and non-future based tasks are compared, what polling-strategy are they using, and what limits?
 
 So the reactive systems that I am referring to usually use stuff like e-poll, kqueue etc. For example if you are using Reactor + Netty there are various system level event driven libraries it can build on you can [read details here](https://netty.io/wiki/native-transports.html). Combined with [project reactor](https://projectreactor.io/), it can get you amazing async DB libraries like [R2DBC](https://r2dbc.io/). 

 > In my mind it should not be pulling in more work before completing a significant part of the last batch. But that will be affected by polling-strategy changes.
 
 If the work it's doing is CPU intensive for sure, even regular threads will be blocked. But if the work is more IO intensive then in traditional model you are just spinning up threads and wasting cycles waiting for it to complete. That's where these 
Async-IO / Reactor driven frameworks let you fully saturate your CPU.

So goin details a little bit more let's assume as simple job scheduler that sends out emails. All the job does is pull an email from worker job and then makes an HTTP call to say SendGrid or some external service to send email. Now in traditional thread model you won't be able to spin up more than couple of hundred thread, so you can't do processing of more than couple of hundred parallel emails at a time, however the async/reactor models unblock you from that. Since each job turns into a future promise based model, none of the threads have to block and wait until the worker lets go of thread. That's what event loops have done and unlocked at large scale. This PR IMO will exactly do that.

kagkarlsson:
I appreciate the explanation, but I was also concretely asking about what settings where used in the test :) 

> In my mind it should not be pulling in more work before completing a significant part of the last batch

was referring to my mental model of how the scheduler works

kagkarlsson:
I think the thing is that if the thread-pool is not the bottleneck in terms of how many executions we may have running in parallel, we need to consider how many is safe to allow? After they finish there will be a backlog of executions to complete..

amit-handda:
Hi @kagkarlsson, we used lock_and_fetch poll strategy with 0.5 (lower_limit) and 4.0 (upper_limit). please let me know if you need more info. Many thanks for your careful review feedback. Have a good day. 

```    
return Scheduler.create(primaryDS, task.make())
      .pollUsingLockAndFetch(lowerLimitFractionOfThreads, upperLimitFractionOfThreads)
      .threads(schConfig.schedulerThreadPool)
      .pollingInterval(Duration.ofSeconds(schConfig.schedulerPollingIntervalSecs.toLong()))
      .schedulerName(SchedulerName.Fixed(schedulerInstanceId))
      .registerShutdownHook()
      .build()
```

perf test config
```
db:
  db:
    jdbcUrl: "jdbc:postgresql://localhost:5432/pgloadtest_local"
    username: "pgloadtest_local"
    password: "testpwd"
    maxLifetimeSecs: 300
    dbconnTtlSecs: 300
    maximumPoolSize: 50
pgloadtest:
  dbscheduler:
    inactivityDurationSecs: 3600
    failRedoSecs: 4
    schedulerThreadPool: 300
    schedulerPollingIntervalSecs: 10
  task:
    failRedoSecs: 5
    inactivityDurationSecs: 600

```

maxpert:
Any updates on if this can be merged? This will really unblock a lot of use-cases.

kagkarlsson:
It is taking some time because it is not a priority feature (though I am starting to see the merits) and there are just still things to consider that I feel are unanswered.

My thoughts atm
* have you verified your fork works as expected for your use-case? 
    - what configuration are you hoping to run?
    - what performance are you hoping to see?
* unexpectedly high executions/s, why is that, it should do maximum 4xthreads/s but seem to be doing 5-6xthreads/s
  - was the same config for comparing sync vs async?
  - do we have a bug in the polling-logic? it should not pull more work until queue-length reaches 0.5xthreads again
* does db-scheduler have any assumptions somewhere in the codebase that the execution lifecycle will be executed by the same thread? (I think not, though aync does make things harder to debug)
* if we go the async route, how would we do stuff like this: https://github.com/kagkarlsson/db-scheduler/issues/116

Also concerned about unknown unknowns. If it goes in, I think it needs to do so as an experimental feature.

amit-handda:
@kagkarlsson 
* digging deeper into ```unexpectedly high executions/s```, ll get back to you
* about async-mod via config setting ... not sure about it, as discussed earlier. the PR makes breaking changes to db-scheduler public API. am open to ideas though.

kagkarlsson:
I feel conflicted about this change. It will enable higher throughput using fewer threads (for non-blocking io use-cases), which is great. On the other hand, I worry about the things it might break or make more complicated. 
And this is why an explicit opt-in for async-mode would be good..

I was considering if `ExecutionHandler` could get a default method for the async-variant, that just wraps the current method in a CompletableFuture. You would override the default method when necessary.. Still, we would need to find a way to toggle the completablefuture code in the executors aswell..

kagkarlsson:
Maybe it is possible to let the async version live and be released from a custom branch 🤔

amit-handda:
Thanks @kagkarlsson I agree with this, we can let it breathe through custom branch please.

maxpert:
If we can do that and release it a some sort of beta package. I can write some reactor benchmarks as well.

amit-handda:
@kagkarlsson happy friday, checking in. May I help getting it released from custom branch ? thanks

kagkarlsson:
Shall we recreate the PR against the `async` branch I created?

kagkarlsson:
Trying to merge into the `async` branch now, but I am getting compilation errors. Is this branch building ok?

GithubRyze:
I reviewed the past comment about this issue and curious also.    we can do like below code to impl non blokc task(like http request).  and on the other hand task is async or not it depends on use-case
```
 public RecurringTaskWithPersistentSchedule<PlainScheduleAndData> asyncRecurringTask() {
    return Tasks
        .recurringWithPersistentSchedule("async-recurring-task", PlainScheduleAndData.class)
          CompletableFuture<String> orgFuture = CompletableFuture.supplyAsync(
              ()->{
                try {
                  // mock http request
                  Thread.sleep(2000L);
                } catch (InterruptedException e) {
                  e.printStackTrace();
                }
              return "test";
              }
          );
          orgFuture.whenComplete((a, throwable) -> {
            // call back here

          });
        });
  }

```

kagkarlsson:
Sorry for not having the bandwidth for pursuing this issue. What about using virtual threads to solve this?

maxpert:
@kagkarlsson virtual thread are in EAP and even after rollout just like record classes it's gonna take years for people to move over. All of existing library and software stack still relies on `Future` based solutions.

kagkarlsson:
I am experimenting some with this feature in #369. Copied the ideas from this PR, but wanted a way to experiment without affecting existing code and tasks, so making it a slightly more hack-ish add-on for now.
I see you guys have a lot of experience with non-blocking code, feel free to pitch in there @amit-handda @maxpert 

https://github.com/kagkarlsson/db-scheduler/pull/369

See example [AsyncOneTimeTaskMain.java](https://github.com/kagkarlsson/db-scheduler/blob/619c7226f2f4fc4b9e1fa7dc3580350620ec15df/examples/features/src/main/java/com/github/kagkarlsson/examples/AsyncOneTimeTaskMain.java)

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/SchedulerClientTest.java
```diff
@@ -65,23 +66,27 @@ public void setUp() {
...
 
     @Test
-    public void client_should_be_able_to_schedule_executions() {
+    public void client_should_be_able_to_schedule_executions() throws InterruptedException {
         SchedulerClient client = SchedulerClient.Builder.create(DB.getDataSource()).build();
         client.schedule(oneTimeTaskA.instance("1"), settableClock.now());
 
         scheduler.runAnyDueExecutions();
+        // Since execution is executed in an async way, we need to wait for a while to let the execution finish before asserting
+        Thread.sleep(1000);
```
kagkarlsson, line 75: Sleeping in tests is an anti-pattern I am trying to avoid as much as possible :)

matt-snider:
> What about using virtual threads to solve this?

@kagkarlsson Virtual threads would probably be much much simpler to support for users who can use it (starting from Java 19 with `--enable-preview`). This might just work without issue, but I haven't tried it. 

```java
Scheduler.create(...)
  // ... 
  .executorService(Executors.newVirtualThreadPerTaskExecutor())
  .build();
```

If I get a chance to experiment with this I will report back

ml-deb:
> > What about using virtual threads to solve this?
> 
> @kagkarlsson Virtual threads would probably be much much simpler to support for users who can use it (starting from Java 19 with `--enable-preview`). This might just work without issue, but I haven't tried it.
> 
> ```java
> Scheduler.create(...)
>   // ... 
>   .executorService(Executors.newVirtualThreadPerTaskExecutor())
>   .build();
> ```
> 
> If I get a chance to experiment with this I will report back

I'm also interested on knowing your results with virtual threads.

kagkarlsson:
I will create a separate issue for virtual-threads support instead of CompletableFuture. Closing this PR.

Issue #298, opened 2022-06-25
Add a new ExecutionOperation, removeAndScheduleNew


kagkarlsson:
Add a new `ExecutionOperations.removeAndScheduleNew` method to allow removing current execution and schedule a completely new one (all values reset to defaults) in a singular statement. (More efficient than deleting the current one and inserting a new record.

Will enable job-chaining using CompletionHandlers without the need for transactional CompletionHandlers

kagkarlsson:
Fixed in v11.4

PR #319, opened 2022-09-13
Add support for ExecutionOperations.replace

kagkarlsson:
Enables JobChaining by replacing the old execution with the next step in the CompletionHandler.


kagkarlsson:
Fixes #298

PR #329, opened 2022-10-03
Allow setting onFailure for RecurringTaskWithPersistentSchedule

olayinkasf:


kagkarlsson:
Makes sense. What is your particular need?

olayinkasf:
The changes proposed:

- Allow setting onFailure when building RecurringTaskWithPersistentSchedule.
- Allow supplying just onFailure to the constructors. I have about 14 task classes all inheriting from different tasks and having different failure handlers. When inheriting from the tasks, I had to supply both the onFailure  and dead execution handler which is inconvenient given that, all 14 task classes revive dead executions. Having the opportunity to supply just the onFailure in the constructor reduces the additional noisy code.
- Allow setting the TaskResolver via the builder. It is currently not possible to handle the case where a task depends on the scheduler. Having the TaskResolver independent of the scheduler allows us to add and remove tasks any time we want especially after building the scheduler.
- Allow setting the clock via the builder which allows the scheduler to use the same clock as the rest of the app. Creating the waiter which depends on the clock and the polling interval is then deferred until the build method is called.

kagkarlsson:
Will have a look at this on Friday

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerBuilder.java
```diff
@@ -91,8 +92,18 @@ public <T extends Task<?> & OnStartup> SchedulerBuilder startTasks(List<T> start
         return this;
     }
 
+    public SchedulerBuilder taskResolver(TaskResolver taskResolver) {
+        this.taskResolver = taskResolver;
+        return this;
+    }
```
kagkarlsson, line 98: I am uncertain what the use-case for this is. If you do not use the default or a variant of it not much will work

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerBuilder.java
```diff
@@ -91,8 +92,18 @@ public <T extends Task<?> & OnStartup> SchedulerBuilder startTasks(List<T> start
...
+
+    public SchedulerBuilder clock(Clock clock) {
+        this.clock = clock;
+        return this;
+    }
+
```
kagkarlsson, line 104: What would be the use-case for this?

kagkarlsson:
> It's only a tiny change. There are a few other tiny improvements observed and I'm not certain what the recommendations are. Batch them in a single PR or separate in multiple PRs?

I think split out task-resolver and clock stuff into a separate PR, then this PR will be fine. I might merge it and remove those bits, to be able to merge the FailureHandler change today.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerBuilder.java
kagkarlsson: I see you have commented a bit on it. We can take the discussion in another PR

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerBuilder.java
kagkarlsson: I see you have commented a bit on it. We can take the discussion in another PR

kagkarlsson (approved):
I will merge this PR, but remove the clock, task-resolver change. I think those changes will block the merge otherwise, and was hoping to release a new version today

kagkarlsson:
I think we need to discuss this: 

> Allow setting the TaskResolver via the builder. It is currently not possible to handle the case where a task depends on the scheduler. Having the TaskResolver independent of the scheduler allows us to add and remove tasks any time we want especially after building the scheduler.

To avoid circular dependency, you can get a `SchedulerClient` from `executionContext.getSchedulerClient()` in the ExecutionHandler

Issue #336, opened 2022-10-28
More Spring Boot examples


kagkarlsson:
There are a lot of confusion around how different features should be used when using the spring boot starter. Need some more examples for advanced use-cases with spring

dipeshsingh253:
Hey @kagkarlsson , I would like to contribute. I am good with spring, spring boot and other related tools.

kagkarlsson:
This issue is a bit out-dated, it was mostly fixed by introducing more examples here: https://github.com/kagkarlsson/db-scheduler/tree/master/examples/spring-boot-example

Appreciate the offer to contribute though 🙏. Do you see any other interesting candidates for contribution?

dipeshsingh253:
Thanks for the update @kagkarlsson, I will create an issue in case I find one or see any room for improvement.

kagkarlsson:
🎉 This issue has been resolved in `v15.1.2` ([Release Notes](https://github.com/kagkarlsson/db-scheduler/releases/tag/v15.1.2))

PR #349, opened 2022-12-30
Spring examples

kagkarlsson:
## More comprehensive Spring Boot examples

Addresses #336

## Reminders
- ✅  Added/ran automated tests

### DIFF
--- a/db-scheduler/pom.xml
+++ b/db-scheduler/pom.xml
@@ -76,6 +76,11 @@
             <artifactId>jackson-annotations</artifactId>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-jsr310</artifactId>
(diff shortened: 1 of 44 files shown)


lucarota:
Hi, when, if ever, will this PR be merged?

kagkarlsson:
It is the current one I work on, so hopefully soon

michalhosala:
Thanks for your work and the examples. However, I am failing to see what would be the best way to schedule one time task in springboot web app from these examples and would appreciate some more info on that. I created one time task bean pretty much like [here](https://github.com/kagkarlsson/db-scheduler/blob/master/examples/spring-boot-example/src/main/java/com/github/kagkarlsson/examples/boot/config/BasicExamplesConfiguration.java#L75), but no idea where to schedule the task in the springboot web app code.

With recurring tasks there is no such issue as I specify `fixedDelay` and they are scheduled without any additional code, but it is not clear at the first sight where to schedule one time tasks under such circumstances.

PR #357, opened 2023-02-02
Add support for custom JdbcCustomization via Spring Boot

### LINKED ISSUE #354: MSSQL wrong data type for task_data column
# Prerequisites

Please answer the following questions for yourself before submitting an issue. **YOU MAY DELETE THE PREREQUISITES SECTION.**

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

## Expected Behavior

Serialization/deserialization of task data works in MSSQL

## Current Behavior

Serialization/deserialization of task data does not work in MSSQL, because of wrong data type. According to documentation nvarbinary should be used to store binary data, however nvarchar is used.
https://learn.microsoft.com/en-us/sql/t-sql/data-types/binary-and-varbinary-transact-sql?redirectedfrom=MSDN&view=sql-server-ver16. After changing the column data type, serialization/deserialization works as expected.

--------
# For bug reports
**YOU MAY DELETE THE `For bug reports` SECTION IF A NEW FEATURE REQUEST.**

## Steps to Reproduce the bug
1. use MSSQL as a persistent storage
2. create task with any data
3. wait for task to be triggered

## Context
- DB-Scheduler Version : 11.6
- Java Version : 17,0,6
- Database and Version : mcr.microsoft.com/mssql/server:2019-latest

quaso:
Column task_data uses correct data type (for binary data) in MSSQL according to MS documentation

## Fixes
#354 

## Reminders

---
cc @kagkarlsson


quaso:
@kagkarlsson can you please review this PR. The bug is blocking us o:-)

kagkarlsson:
Currently on vacation for 1w, so cant look at it atm

kagkarlsson:
Sure this will not break for schemas of the old type?

quaso:
I just tested - this change is not backward compatible. The question is if anybody is using task_data column in mssql, because from what I can see it cannot work currently.

quaso:
The problem seems to occur for MSSQL 2019+

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
```diff
@@ -103,7 +101,7 @@ public boolean createIfNotExists(SchedulableInstance instance) {
                     (PreparedStatement p) -> {
                         p.setString(1, taskInstance.getTaskName());
                         p.setString(2, taskInstance.getId());
-                        p.setObject(3, serializer.serialize(taskInstance.getData()));
+                        p.setBytes(3, serializer.serialize(taskInstance.getData()));
```
kagkarlsson, line 104: It will not work without this change?

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/MssqlJdbcCustomization.java
```diff
@@ -44,6 +51,27 @@ public Instant getInstant(ResultSet rs, String columnName) throws SQLException {
...
+                while ((numCharsRead = reader.read(charArray, 0, charArray.length)) != -1) {
+                    builder.append(charArray, 0, numCharsRead);
+                }
+                result = builder.toString().getBytes(StandardCharsets.UTF_16LE);
+            } catch (IOException e) {
+                throw new SQLException(e);
+            }
+        } else {
+            result = rs.getBytes(columnName);
+        }
```
kagkarlsson, line 71: How about making this into a separate JdbcCustomization, e.g. `MssqlJTDSJdbcCustomization`?

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcCustomization.java
```diff
@@ -29,6 +29,10 @@ public interface JdbcCustomization {
     void setInstant(PreparedStatement p, int index, Instant value) throws SQLException;
     Instant getInstant(ResultSet rs, String columnName) throws SQLException;
 
+    default byte[] getBytes(ResultSet rs, String columnName) throws SQLException {
+        return rs.getBytes(columnName);
+    }
+
```
kagkarlsson, line 35: Good stuff moving the `getBytes(..)` into the Customization

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/AutodetectJdbcCustomization.java
```diff
@@ -44,7 +44,7 @@ public AutodetectJdbcCustomization(DataSource dataSource) {
 
             if (databaseProductName.equals(MICROSOFT_SQL_SERVER)) {
                 LOG.info("Using MSSQL jdbc-overrides.");
-                detectedCustomization = new MssqlJdbcCustomization();
+                detectedCustomization = new MssqlJdbcCustomization(c.getMetaData().getDriverName());
```
kagkarlsson, line 47: As mentioned below I think the jTDS should be a separate Customization, and hence no need to change standard `MssqlJdbcCustomization`

Is the last release of jTDS from 2013??

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/CronSchedule.java
```diff
@@ -39,6 +39,8 @@ public class CronSchedule implements Schedule, Serializable {
 
     private static final String DISABLED = "-";
     private static final Logger LOG = LoggerFactory.getLogger(CronSchedule.class);
+
+    private static final long serialVersionUID = 6577185347687546367L;
```
kagkarlsson, line 43: Did you need this?

quaso: Not sure why, but I receive InvalidClassException, if this ID is missing and database entry was created with master version. I used the same JVM for compiling master and this branch, but still the error occurs - not sure why.

```
Caused by: java.io.InvalidClassException: com.github.kagkarlsson.scheduler.task.schedule.CronSchedule; local class incompatible: stream classdesc serialVersionUID = 6577185347687546367, local class serialVersionUID = 6577185347687514335
	at java.base/java.io.ObjectStreamClass.initNonProxy(ObjectStreamClass.java:597) ~[na:na]
	at java.base/java.io.ObjectInputStream.readNonProxyDesc(ObjectInputStream.java:2051) ~[na:na]
	at java.base/java.io.ObjectInputStream.readClassDesc(ObjectInputStream.java:1898) ~[na:na]
	at java.base/java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2224) ~[na:na]
	at java.base/java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1733) ~[na:na]
	at java.base/java.io.ObjectInputStream$FieldValues.<init>(ObjectInputStream.java:2606) ~[na:na]
	at java.base/java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:2457) ~[na:na]
	at java.base/java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2257) ~[na:na]
	at java.base/java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1733) ~[na:na]
	at java.base/java.io.ObjectInputStream.readObject(ObjectInputStream.java:509) ~[na:na]
	at java.base/java.io.ObjectInputStream.readObject(ObjectInputStream.java:467) ~[na:na]
	at com.github.kagkarlsson.scheduler.boot.autoconfigure.DbSchedulerAutoConfiguration$2.deserialize(DbSchedulerAutoConfiguration.java:232) ~[db-scheduler-spring-boot-starter-master-SNAPSHOT.jar:na]
```

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/AutodetectJdbcCustomization.java
quaso: Yes, the latest version was released in 2013

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/MssqlJdbcCustomization.java
quaso: I created new JdbcCustomization

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/CronSchedule.java
kagkarlsson: Hmm, that's a bit strange. When I think about it, I suppose the `serialVersionUID` really should be set in db-scheduler for everything that might be serialized.

kagkarlsson: I think you need to investigate that a bit more. We cannot add that without reason since it might break serialization for other users. That said, it should really have a `serialVersionUID` set..

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/MssqlJtdsJdbcCustomization.java
```diff
@@ -0,0 +1,48 @@
...
+            while ((numCharsRead = reader.read(charArray, 0, charArray.length)) != -1) {
+                builder.append(charArray, 0, numCharsRead);
+            }
+            result = builder.toString().getBytes(StandardCharsets.UTF_16LE);
+        } catch (IOException e) {
+            throw new SQLException(e);
+        }
+        return result;
+    }
+}
```
kagkarlsson, line 48: Will you not need a similar override for writing the bytes? It looks a bit suspicious specifying charset only one way (`StandardCharsets.UTF_16LE`)

```java
p.setObject(3, serializer.serialize(taskInstance.getData()));
```

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/schedule/CronSchedule.java
kagkarlsson: It was a mistake on my part not setting a value for `serialVersionUID` for this class. Serialization will break for anyone using this class when we set it. The solution will be to migrate away from the class temporarily.

kagkarlsson:
I pushed some changes to your branch enabling configuring the `JdbcCustomization` in spring boot starter as well. TBH I think `MssqlJtdsJdbcCustomization` should live on your end and set via 

```java
@Bean
    DbSchedulerCustomizer customizer() {
        ....
```

How does that sound?

kagkarlsson:
I pushed some changes to your branch enabling configuring the `JdbcCustomization` in spring boot starter as well. TBH I think `MssqlJtdsJdbcCustomization` should live on your end and set via 

```java
@Bean
    DbSchedulerCustomizer customizer() {
        ....
```

How does that sound?

kagkarlsson:
I will likely merge this PR, but remove `MssqlJtdsJdbcCustomization` from the codebase. You will need to move that to your own code and set it explicitly via `DbSchedulerCustomizer`

quaso:
Seems like your commit does the trick and there is no need for my code changes in this repo - my commit might be dropped. Or do I miss something?

Issue #348, opened 2022-12-20
Database Deadlocks with MSSQL 2019


rafaelhofmann:
# Prerequisites
- [ ] I am running the latest version (We are running 11.5)
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

--------
# For bug reports
When we execute around 300 - 400 tasks per minute, we experience database deadlocks on the table `scheduled_tasks`.
The deadlock report from the MSSQL database shows a deadlock between either one of the indexes `execution_time_idx` or `last_heartbeat_idx` and the clustered primary key of the table.

The deadlock happens when a Custom-Task tries to reschedule itself for another execution (`PollingTaskConfiguration:42` in the stacktrace below) with `executionOperations.reschedule`.

## Workaround
For now we have deleted both indexes to stop the deadlocks from happening (so far successfully), but probably at the cost of reduced throughput.

## Steps to Reproduce the bug
Currently unclear

## Context
- DB-Scheduler Version : 11.5
- Java Version : 1.8
- Spring Boot (check for Yes) : [x]
- Database and Version : Microsoft SQL Server 2019
- Database Transaction Isolation Level: Read Committed Snapshot

## Logs
```

Failed while completing execution Execution: task=polling-task, id=polling-task-155175, executionTime=2022-12-17T23:28:53.936Z, picked=true, pickedBy=workflow-engine-36-v6vm9, lastHeartbeat=2022-12-17T23:29:03.339Z, version=642. Execution will likely remain scheduled and locked/picked. The execution should be detected as dead after a while, and handled according to the tasks DeadExecutionHandler.
                com.github.kagkarlsson.jdbc.SQLRuntimeException: com.microsoft.sqlserver.jdbc.SQLServerException: Transaction (Process ID 71) was deadlocked on lock resources with another process and has been chosen as the deadlock victim. Rerun the transaction.
                at com.github.kagkarlsson.jdbc.JdbcRunner.translateException(JdbcRunner.java:126)
                at com.github.kagkarlsson.jdbc.JdbcRunner.lambda$execute$2(JdbcRunner.java:93)
                at com.github.kagkarlsson.jdbc.JdbcRunner.withConnection(JdbcRunner.java:140)
                at com.github.kagkarlsson.jdbc.JdbcRunner.execute(JdbcRunner.java:66)
                at com.github.kagkarlsson.jdbc.JdbcRunner.execute(JdbcRunner.java:54)
[...]

Issue #337, opened 2022-11-01
Is there a plan to support MSSQL for lock_and_fetch. Using fetch polling strategy has performance issue.


AnanthaKrishnaV:
We are using 11.5 version and facing performance issues. We are using MSSQL DB, we have 10 pod running with 8 threads configured, 10*8 = 80 threads total All other configuration are left to default.  Scheduler is able to process only 60 to 80 transaction per hour.  For our business each transaction take ~40 to 50 secs. Any suggestion to improve the throughput with 'fetch' polling would really help.

kagkarlsson:
Are there much data in the database?

AnanthaKrishnaV:
> Are there much data in the database?

Yes, we have around 1.5M records in the scheduler_task table.

kagkarlsson:
You might need an index. For example for postgres: 

```
CREATE INDEX execution_time_idx ON scheduled_tasks (execution_time);
```

kagkarlsson:
You probably need to try and debug and see what is happening

Issue #273, opened 2022-03-17
MSSQL and snaphost isolation issues


cezarg1410:
HI,
I am using db-scheduler 10.5 version with Spring Boot 2.5.6.
Database which i use is Microsoft SQL Server 2019.
I use db scheduler autoconfiguration so basically the whole configuration of db-scheduler is just:

```
db-scheduler:
  threads: 5
  heartbeat-interval: 1m
  polling-interval: 300ms
  delay-startup-until-context-read: true
```

What's really important: for some other reasons we HAD to switch to Snaphost isolation in MSSQL. Unfortunately Spring Boot doesn't support that isolation so it had to be set on hikari connection level:

```
spring:
    hikari:
      transaction-isolation: 4096 #That means SNAPSHOT isolation - can't be set using Spring Boot
```

The problem is: from time to time i can see the following errors in the logs:
```
Exception in thread "*****************" com.github.kagkarlsson.jdbc.SQLRuntimeException: com.microsoft.sqlserver.jdbc.SQLServerException: Snapshot isolation transaction aborted due to update conflict. You cannot use snapshot isolation to access table 'dbo.scheduled_tasks' directly or indirectly in database '**************************' to update, delete, or insert the row that has been modified or deleted by another transaction. Retry the transaction or change the isolation level for the update/delete statement.
	at com.github.kagkarlsson.jdbc.JdbcRunner.translateException(JdbcRunner.java:126)
	at com.github.kagkarlsson.jdbc.JdbcRunner.lambda$execute$2(JdbcRunner.java:93)
	at com.github.kagkarlsson.jdbc.JdbcRunner.withConnection(JdbcRunner.java:140)
	at com.github.kagkarlsson.jdbc.JdbcRunner.execute(JdbcRunner.java:66)
	at com.github.kagkarlsson.jdbc.JdbcRunner.execute(JdbcRunner.java:54)
	at com.github.kagkarlsson.scheduler.jdbc.JdbcTaskRepository.pick(JdbcTaskRepository.java:234)
	at com.github.kagkarlsson.scheduler.FetchCandidates$PickDue.call(FetchCandidates.java:120)
	at com.github.kagkarlsson.scheduler.FetchCandidates.lambda$run$2(FetchCandidates.java:84)
	at com.github.kagkarlsson.scheduler.Executor.lambda$addToQueue$0(Executor.java:52)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
[...]

Issue #264, opened 2022-02-15
Polling - support select-for-update-skip-locked polling for mysql, mssql


kagkarlsson:

kagkarlsson:
Groundwork laid, but not prioritizing this anymore.

kagkarlsson:
🎉 This issue has been resolved in `v15.1.0` ([Release Notes](https://github.com/kagkarlsson/db-scheduler/releases/tag/v15.1.0))

PR #371, opened 2023-04-17
feat: Support lock-and-fetch for MSSQL. Explicit LIMIT for all.

kagkarlsson:
* Adding support for `lock-and-fetch` aka. `SELECT FOR UPDATE .. SKIP LOCKED` for MSSQL/SqlServer. Though testing has shown that lock-and-fetch are prone to deadlocks, so it is not recommended until that is understood/resolved. However, a query-hint was also added to `fetch-and-lock-on-execute` for MSSQL, and it appears this resolves issues with deadlocks for that strategy. 
* Updates jdbc-handling dependency `micro-jdbc`
* Re-written transaction-handling (in `micro-jdbc`)
* Shading `micro-jdbc` dependency
* Explicit limit also for SqlServer, MySQL, Oracle
* Adds ClusterTest for Sql Server (test concurrency)

## Fixes

* #264
* #337
* #348 ?
* #273 ?

## Further work / fix later

* Mysql v8 skip locked syntax
* MariaDB skip locked syntax
* Oracle skip locked syntax

## Reminders

---
cc @kagkarlsson


kagkarlsson:
This is a bit tricky to test since Testcontainers does not seem to work on M1 Macs and testing via Github Actions sporadically fails. I have done some local testing using AzureSQLEdge, and the lock-and-fetch strategy _seem_ to be working. If anyone using MSSQL could try this branch out that would be helpful. 

With a little luck this might fix the deadlock issues experienced using the default polling strategy. I have gotten the deadlock locally (AzureSQLEdge) using default polling, but not using lock-and-fetch.

Is it possible for you do help out verifying this change for MSSQL @rafaelhofmann ?


This is the query currently in use:

```java
    @Override
    public String createGenericSelectForUpdateQuery(String tableName, int limit, String requiredAndCondition) {
        return "select TOP "+limit+" * from " + tableName +
            " WITH (readpast,rowlock) where picked = ? and execution_time <= ? " + requiredAndCondition +
            " order by execution_time asc ";
    }
```

kagkarlsson:
I added `with (readpast)` to the regular query as well for mssql (locally), and now I do not get the dead-lock issues @rafaelhofmann

rafaelhofmann:
Hi @kagkarlsson 

I will test this new branch and check if we still have problems with the indexes.

rafaelhofmann:
Hi @kagkarlsson,

Do you have a published snapshot version that we can use to test or should we build our own version?

Thanks

kagkarlsson:
Currently you would need to build your own. Let me know if that is a problem.

rafaelhofmann:
I have built the branch and upgraded our ACPT system (including all the indexes that were deleted in the past) to the new release, but we are running into the same problem as before: A deadlock about once every hour, when we have a load of 300 tasks/min.
I have been experimenting with a query that combines the SELECT and UPDATE in one, but I am not quite happy yet with the query hint for MSSQL:
```java
String selectForUpdateQuery =
            " UPDATE " + ctx.tableName +
                " SET " + ctx.tableName + ".picked = ?, " +
                "     " + ctx.tableName + ".picked_by = ?, " +
                "     " + ctx.tableName + ".last_heartbeat = ?, " +
                "     " + ctx.tableName + ".version = " + ctx.tableName + ".version + 1 " +
                " OUTPUT [inserted].* " +
                " FROM ( " +
                "   SELECT TOP(?) ist2.task_name, ist2.task_instance " +
                "   FROM " + ctx.tableName + " ist2 WITH (ROWLOCK, READPAST) " +
                "   WHERE ist2.picked = ? AND ist2.execution_time <= ? " + unresolvedFilter.andCondition() +
                "   ORDER BY ist2.execution_time ASC " +
                " ) AS st2 " +
                " WHERE st2.task_name = " + ctx.tableName + ".task_name " +
                "   AND st2.task_instance = " + ctx.tableName + ".task_instance";
```

In theory this should do the same as the `lock-and-fetch` query from PostgreSQL.

kagkarlsson:
Ok, thanks. Did you try both "lock-and-fetch" and "fetch-and-lock-on-execute"? I think they should both work for MSSQL now.

kagkarlsson:
Your combined query looks interesting

rafaelhofmann:
We had deadlocks with the `lock-and-fetch` strategy, but so far no errors with the default `fetch` strategy after four hours of running. Will keep this running for a few more days to verify.

kagkarlsson:
How has the `fetch` strategy run?

rafaelhofmann:
The `fetch` strategy runs without any deadlock or error for the last 30 days with all the recommended indexes.

kagkarlsson:
Ok, that is good news. I was hoping for good news on the other strategy as well, but I will take 1 out of 2 :).

This means you are happy with the changes in the PR and would like to have it merged?

kagkarlsson:
Good question, nothing blocking it from being released, just time I guess :). Hopefully I can release it within 1-2 weeks

ChristianCiach:
Hi! I am currently evaluating db-scheduler for our use-cases and doing some code reviews in the process. 

I see that the `MariaDBJdbcCustomization`, now uses `FOR UPDATE SKIP LOCKED` for SELECT-queries even though MariaDB only supports `SKIP LOCKED` since MariaDB 10.6, according to the documentation: https://mariadb.com/kb/en/select/#skip-locked

I don't particularly care about older versions of MariaDB, but I just want to let you know that the versions 10.4 and 10.5 are technically still supported, according to https://mariadb.com/kb/en/mariadb-server-release-dates/

Anyway, thank you for your work! I like the simplicity of db-scheduler and I hope we can use it to eventually replace Quartz, which seems to be completely dead by now.

kagkarlsson:
I had some issue when testing the support in MySQL (or MariaDB) and limited time, so I opted for disabling the support until someone had the time to verify it works as expected. Feel free to fork and try it out. 

```java
  @Override
  public boolean supportsGenericLockAndFetch() {
    // FIXLATER: fix syntax and enable it for versions of MariaDB that supports it
    return false;
  }
  ```

PR #418, opened 2023-08-25
fix: getScheduledExecutions(...) should be able to return executions with unresolved tasks

kagkarlsson:
If the `SchedulerClient` is instantiated with no _known tasks_, it currently does not find any executions since the client will not be able to resolve the task and will therefore filter them out. 

This PR **changes the behavior** of `getScheduledExecution()`, `getScheduledExecutions()` to also return unresolved executions, and additionally adds options in `ScheduledExecutionsFilter` to enable controlling whether or not to include them.


kagkarlsson:
🎉 This issue has been resolved in `v12.5.0` ([Release Notes](https://github.com/kagkarlsson/db-scheduler/releases/tag/v12.5.0))

PR #430, opened 2023-10-14
feat: spring-boot: introduce DbSchedulerStopper

### LINKED ISSUE #423: Provide DbSchedulerStopper to control the shutdown of the Scheduler
Hi,

This issue is probably similar to #419, but I would like to ask for a more specific means to control the shutdown.

Currently there is already a _DbSchedulerStarter_ bean definition (which can be "overriden") to control when exactly the _Scheduler_ instance will be started. The shutdown, on the other hand, is not configurable and is perfomed by the _destroyMethod_ at the bean definition within _DbSchedulerAutoConfiguration_. The only possibility to control the shutdown behaviour is by "overriding" the whole bean definition in a configuration class before the _DbSchedulerAutoConfiguration_ and delegating the creation:

```
@ConditionalOnBean(DataSource.class)
@ConditionalOnMissingBean
@DependsOnDatabaseInitialization
@Bean
public Scheduler scheduler(DbSchedulerCustomizer customizer, StatsRegistry registry) {
  return new DbSchedulerAutoConfiguration(...).scheduler(customizer, registry);
}
```

This is rather ugly. It would be very nice, if we could control the shutdown of the _Scheduler_ with a component similar to the _DbSchedulerStarter_. For the _DbSchedulerStopper_ (name pending) a default implementation could react on the event for the closing of the context or something similar.

I would like to detail our motivation for this. In our architecture we have explicit start and shutdown phases, which means we have to control for various components when exactly they are started and stopped. The startup of the _Scheduler_ can easily be controlled, as we are using a _DbSchedulerStarter_ implementation which simply does nothing. For the shutdown, on the other hand, we have to rely on the code above just to "unregister" the destroy method.

## Expected Behavior

Make the shutdown behaviour configurable within a Spring context.

## Current Behavior

Within a Spring context, the shutdown is always performed due to the _destroy_ method.

## Context
- DB-Scheduler Version : 12.5.0
- Java Version : 17

evenh:
This lets the user control when the scheduler should be stopped. By default it stops the scheduler when Spring issues the ContextClosedEvent.

## Fixes
#423 

## Reminders

---
cc @kagkarlsson

### DIFF
--- a/db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/config/shutdown/ContextClosedStopper.java
+++ b/db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/config/shutdown/ContextClosedStopper.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright (C) Gustav Karlsson
+ *
+ * <p>Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * <p>http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * <p>Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.kagkarlsson.scheduler.boot.config.shutdown;
+
+import com.github.kagkarlsson.scheduler.Scheduler;
+import org.springframework.context.event.ContextClosedEvent;
+import org.springframework.context.event.EventListener;
+
+public class ContextClosedStopper extends AbstractSchedulerStopper {
+  public ContextClosedStopper(Scheduler scheduler) {
+    super(scheduler);
+  }
+
+  @EventListener(ContextClosedEvent.class)
+  public void whenContextIsStopped() {
+    doStop();
+  }
+}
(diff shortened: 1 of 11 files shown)


evenh:
Does this fit the requirements @nils-christian?

db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/config/shutdown/ContextClosedStopper.java
```diff
@@ -0,0 +1,29 @@
...
+public class ContextClosedStopper extends AbstractSchedulerStopper {
+  public ContextClosedStopper(Scheduler scheduler) {
+    super(scheduler);
+  }
+
+  @EventListener(ContextClosedEvent.class)
+  public void whenContextIsStopped() {
+    doStop();
+  }
+}
```
kagkarlsson, line 29: Do we know when this event fires? Is it after all beans have been "destroyed"? Are we risking that some of the dependencies of db-scheduler is shutdown before the scheduler? 🤔

pom.xml
```diff
@@ -268,15 +268,15 @@
...
 						<includes>
 							<include>src/main/java/**/*.java</include> <!-- Check application code -->
 							<include>src/test/java/**/*.java</include> <!-- Check application tests code -->
 						</includes>
 						<googleJavaFormat>
-							<version>1.16.0</version>
+							<version>1.18.1</version>
 							<style>GOOGLE</style>
 						</googleJavaFormat>
 						<importOrder />
```
kagkarlsson, line 282: Thanks! 😄

ChristianCiach:
I've added a related comment at https://github.com/kagkarlsson/db-scheduler/issues/423#issuecomment-1930674047

PR #461, opened 2024-01-30
fix: Timestamp fixes for Oracle, MySQL and MariaDB

kagkarlsson:
* **BREAKING CHANGE:** Since **MySQL** and **MariaDB** does not have a datatype for storing timestamps with time zones, start defaulting to storing them in UTC to avoid issues with DST
* **BREAKING CHANGE:** Change **Oracle** schema to use a datatype that supports time zones


kagkarlsson:
This branch's PR was already halfway merged. See #471 for new PR.

PR #471, opened 2024-02-24
MySQL, MariaDB always store timestamps in UTC

kagkarlsson:
* **BREAKING CHANGE:** Since **MySQL** and **MariaDB** does not have a datatype for storing timestamps with time zones, start defaulting to storing them in UTC to avoid issues with DST
* **BREAKING CHANGE:** Change **Oracle** schema to use a datatype that supports time zones


kagkarlsson:
🎉 This issue has been resolved in `v14.0.0` ([Release Notes](https://github.com/kagkarlsson/db-scheduler/releases/tag/v14.0.0))

rydenius:
As it is not mentioned in PR description, was it intentional that there was also a breaking change for PostgreSQL, with the zero argument constructor of PostgreSqlJdbcCustomization  removed?

kagkarlsson:
Not intentional no, more of a slip. Sometimes difficult to remember everything that should make it into a release-note or pr-description

PR #481, opened 2024-04-18
fix: Revert change in behavior for MariaDB and MySQL with regards to time zone assumptions for persisted timestamps

kagkarlsson:
Change back to default behavior for MySQL and MariaDB with regards to timestamp persistence and assumptions. It is recommended to persist timestamps in UTC for users of MariaDB and MySQL, but db-scheduler will not enforce it as default, since it in reality requires a controlled upgrade. 

Ideally the upgrade-process should be:
1. All instances of the scheduler should be shutdown
2. Timestamps in the database-table migrated to UTC
3. Configuration updated to use `alwaysPersistTimestampInUTC()`
4. Instances started again


kagkarlsson:
🎉 This issue has been resolved in `v14.0.0` ([Release Notes](https://github.com/kagkarlsson/db-scheduler/releases/tag/v14.0.0))

PR #497, opened 2024-05-30
fix: Mention exception in log message on failures

runeflobakk:

A small suggestion for some improved ergonomics when reading log output from db-scheduler in the case of failures.

I always find it effective to include a small "summary" of any caught exception in the actual log message. Often this will allow me to pinpoint the error at hand without needing to look into the stacktrace of the logger event. (But this is still possible of course, as before)

Logs when exceptions propagates all the way back to db-scheduler can typically look like this:
db-scheduler-failure-log

This PR will change the log message to mention the actual exception which was caught in the log message:

> Unhandled exception IllegalArgumentException: 'Illegal character in path at index 9: [replace with URL]' during execution of task with name 'my-task-name'. Treating as failure.

## Fixes

No existing issue, I suspect. Please consider this PR the issue, with a suggested fix, all bundled nicely together :)

## Reminders
- [x] Added/ran automated tests (existing tests have been run)
- [x] Update README and/or examples (no update necessary, I think)

---
cc @kagkarlsson

### DIFF
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/ExceptionUtils.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/ExceptionUtils.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright (C) Gustav Karlsson
+ *
+ * <p>Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


kagkarlsson (approved):
Lets ship it!  🏅

kagkarlsson:
🎉 This issue has been resolved in `v14.0.1` ([Release Notes](https://github.com/kagkarlsson/db-scheduler/releases/tag/v14.0.1))

PR #503, opened 2024-06-18
feat: add missing indexes to the oracle ddl script

iluwa:
## Add missing indexes to the oracle ddl script

Every example for other databases (Postgres, Mssql, Mysql) contains indexes on `execution_time` and `last_heartbeat` while for Oracle DB they are missing.

I also expanded `DbUtiils` method to allow splitting sql statements from an sql file. `jdbcRunner.execute(statement, NOOP)` can run only one statement at a time for Oracle Db.

PS: two tests for oracle db are failing on master, is that expected?

## Reminders
    - no need

---
cc @kagkarlsson


kagkarlsson (approved):
Looks good! Thank you for contributing!

Issue #539, opened 2024-09-19
Function Request: putInstance


beilCrxmarkets:
If SchedulerClient.reschedule() is called with a TaskInstance that does not yet exist, an exception is thrown. You must therefore always check whether there is already a scheduled TaskInstance before you can call reschedule. We would therefore be pleased to have a putInstance method that automatically schedules a TaskInstance if there is none to reschedule. This is particularly convenient for dynamic recurring tasks, as these first have to be initialized. 

```
    void putInstance(TaskInstance<Void> taskInstance, Instant executionTime) {
        var optionalExecutionObject = schedulerClient.getScheduledExecutions()
                .stream()
                .filter(e -> e.getTaskInstance().getId().equals(taskInstance.getId()))
                .findFirst();
        if (optionalExecutionObject.isPresent()) {
            schedulerClient.reschedule(taskInstance, executionTime);
        }
        else {
            schedulerClient.scheduleIfNotExists(taskInstance, executionTime);
        }
    }
```

PR #553, opened 2024-11-12
feat: Add schedule-or-reschedule method to the SchedulerClient

beilCrxmarkets:
## feat: Add rescheduleOrCreate method to the SchedulerClient
When invoking SchedulerClient.reschedule() with a non-existent TaskInstance, an exception will be raised. Therefore, it's essential to verify the existence of a scheduled TaskInstance prior to attempting a reschedule. To streamline this process, we propose the introduction of a scheduleOrCreate method. This method would automatically schedule a TaskInstance if none exists, making it particularly beneficial for dynamically recurring tasks that require initial setup.

We have given considerable thought to finding an appropriate name for this method, but unfortunately, we haven't arrived at a satisfactory conclusion. The verb "schedule" best describes its functionality; however, it cannot be used as there is already a method with that name. Since "rescheduleOrSchedule" sounds awkward, we have settled on "rescheduleOrCreate" for now. We are open to considering other names for this method.

## Fixes
[#539](https://github.com/kagkarlsson/db-scheduler/issues/539)

## Reminders

---
cc @kagkarlsson

### DIFF
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerClient.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerClient.java
@@ -58,7 +58,8 @@ public interface SchedulerClient {
    * @see com.github.kagkarlsson.scheduler.task.TaskInstance
    * @see com.github.kagkarlsson.scheduler.exceptions.TaskInstanceCurrentlyExecutingException
    */
-  <T> boolean schedule(TaskInstance<T> taskInstance, Instant executionTime, ScheduleOptions scheduleOptions);
+  <T> boolean schedule(
+      TaskInstance<T> taskInstance, Instant executionTime, ScheduleOptions scheduleOptions);
 
   /**
    * Schedule a new execution for the given task instance.
@@ -246,14 +247,13 @@ public interface SchedulerClient {
    */
   Optional<ScheduledExecution<Object>> getScheduledExecution(TaskInstanceId taskInstanceId);
 
-
   class ScheduleOptions {
 
     public static final ScheduleOptions WHEN_EXISTS_DO_NOTHING =
         defaultOptions().whenExistsDoNothing();
 
     public static final ScheduleOptions WHEN_EXISTS_RESCHEDULE =
-      defaultOptions().whenExistsReschedule();
+        defaultOptions().whenExistsReschedule();
 
     public enum WhenExists {
       RESCHEDULE,
@@ -418,9 +418,9 @@ public interface SchedulerClient {
         if (existing.isEmpty()) {
           // something must have processed it and deleted it
           LOG.warn(
-            "Task-instance should already exist, but failed to find it. "
-              + "It must have been processed and deleted. task-instance={}",
-            taskInstance);
+              "Task-instance should already exist, but failed to find it. "
+                  + "It must have been processed and deleted. task-instance={}",
+              taskInstance);
           return false;
         }
 
@@ -433,7 +433,8 @@ public interface SchedulerClient {
     }
 
     @Override
-    public <T> boolean schedule(SchedulableInstance<T> schedulableInstance, ScheduleOptions whenExists) {
+    public <T> boolean schedule(
+        SchedulableInstance<T> schedulableInstance, ScheduleOptions whenExists) {
       return schedule(
           schedulableInstance.getTaskInstance(),
           schedulableInstance.getNextExecutionTime(clock.now()),
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerClient.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerClient.java
@@ -15,6 +15,7 @@ package com.github.kagkarlsson.scheduler;
 
 import static java.util.Optional.ofNullable;
 
+import com.github.kagkarlsson.scheduler.SchedulerClient.ScheduleOptions.WhenExists;
 import com.github.kagkarlsson.scheduler.event.SchedulerListeners;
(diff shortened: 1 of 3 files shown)


db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerClient.java
```diff
@@ -352,6 +380,26 @@ public <T> void reschedule(TaskInstanceId taskInstanceId, Instant newExecutionTi
       }
     }
 
+    @Override
+    public <T> void rescheduleOrCreate(TaskInstance<T> taskInstance, Instant executionTime) {
+      Optional<ScheduledExecution<Object>> executionObject =
+          this.getScheduledExecutions().stream()
+              .filter(e -> e.getTaskInstance().getId().equals(taskInstance.getId()))
+              .findFirst();
```
kagkarlsson, line 388: We should avoid fetch all here, instead use
```
Optional<ScheduledExecution<Object>> getScheduledExecution(TaskInstanceId taskInstanceId);
```

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerClient.java
```diff
@@ -352,6 +380,26 @@ public <T> void reschedule(TaskInstanceId taskInstanceId, Instant newExecutionTi
...
+      if (executionObject.isPresent()) {
+        this.reschedule(taskInstance, executionTime, taskInstance.getData());
+      } else {
+        this.scheduleIfNotExists(taskInstance, executionTime);
+      }
```
kagkarlsson, line 393: As long as we do not lock the row, there will be the possibility of race-conditions. 

* Maybe log a warn if `scheduleIfNotExists` returns false
* Also, maybe make `reschedule` return the success-code and warn on false on that as well.

kagkarlsson, line 393: As long as we do not lock the row, there will be the possibility of race-conditions. 

* Maybe log a warn if `scheduleIfNotExists` returns false
* Also, maybe make `reschedule` return the success-code and warn on false on that as well.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerClient.java
```diff
@@ -352,6 +380,26 @@ public <T> void reschedule(TaskInstanceId taskInstanceId, Instant newExecutionTi
...
+    }
+
+    @Override
+    public <T> void rescheduleOrCreate(SchedulableInstance<T> schedulableInstance) {
+      rescheduleOrCreate(
+          schedulableInstance.getTaskInstance(),
+          schedulableInstance.getNextExecutionTime(clock.now()));
+    }
```
kagkarlsson, line 401: Agreed that it is hard to come up with a good name for this operation.. 🤔 

rescheduleOrCreate
scheduleOrReschedule
scheduleOrMove
scheduleOrReplace
schedule(id, WhenExists.RESCHEDULE).  (OVERWRITE/DO_NOTHING...)
..

What do you think about a variant of the last one, where you have a parameter for what to do when the task-instance already exists?

kagkarlsson (commented):
Thank you for contributing! A couple of comments :)

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerClient.java
beilCrxmarkets: You are right about that. I have adopted your suggestion.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerClient.java
beilCrxmarkets: In my opinion, scheduleIfNotExists should generally create a warning log entry in the event of failure. This is already the case. I therefore do not see the need to generate an additional log entry for this case. 

reschedule now also returns a boolean value and also creates a warning log entry in the event of an error. The only question here is how useful this is at the moment, as an ExecutionException is thrown in the only possible error case. It is therefore currently not possible for reschedule to return false, as an exception is thrown in this case.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerClient.java
beilCrxmarkets: schedule as the method name with an additional parameter seems most intuitive to me. 

I have therefore adapted the code accordingly. 
schedule now also returns a boolean value, just like scheduleIfNotExists and reschedule.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerClient.java
```diff
@@ -322,42 +365,77 @@ public <T> void schedule(SchedulableInstance<T> schedulableInstance) {
...
       String instanceId = taskInstanceId.getId();
-      Optional<Execution> execution = taskRepository.getExecution(taskName, instanceId);
-      if (execution.isPresent()) {
-        if (execution.get().isPicked()) {
-          throw new TaskInstanceCurrentlyExecutingException(taskName, instanceId);
-        }
+      Execution execution =
+          taskRepository
+              .getExecution(taskName, instanceId)
+              .orElseThrow(() -> new TaskInstanceNotFoundException(taskName, instanceId));
```
beilCrxmarkets, line 420: With the use of `orElseThrow` we can get rid of the big brackets which goes around 10+ lines of code, which should improve the code readability.

kagkarlsson (approved):
Looks good 👍

Issue #538, opened 2024-09-18
MaxRetriesFailureHandler does not work with OnFailureReschedule


beilCrxmarkets:
## Configuration
```
@Configuration
public class DbSchedulerConfiguration {

    @Bean
    RecurringTask<Void> staticRecurringTask() {
        CronSchedule schedule = cron("*/1 * * * * *"); // High frequency for testing purposes. Normally once a day. 

        return Tasks.recurring("Trigger job", schedule)
                .onFailure(new FailureHandler.MaxRetriesFailureHandler<>(5,
                        new FailureHandler.ExponentialBackoffFailureHandler<>(Duration.ofSeconds(3), 1),
                        new FailureHandler.OnFailureReschedule<Void>(schedule)::onFailure))
                .execute((instance, ctx) -> {
                    System.out.println("Running staticRecurringTask");
                    throw new RuntimeException("Error");
                });
    }
}
```

## Expected Behavior
"Trigger job" is rescheduled again after the 5 retry attempts have failed. 

## Current Behavior
"Trigger job" has been removed from the DB by calling `executionOperations.stop();` See: https://github.com/kagkarlsson/db-scheduler/blob/master/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/FailureHandler.java#L96

Therefore the attempt to reschedule the task will fail. 

## Proposed solution

If a maxRetriesExceededHandler is defined, let it decide whether the task is stopped/removed. 

--------
## Context
- DB-Scheduler Version : 14.0.3
- Java Version : 21
- Spring Boot (check for Yes) : [X]
- Database and Version : PostgreSQL

## Logs
Running staticRecurringTask
[...]

PR #554, opened 2024-11-12
feat: Improve flexibility of the MaxRetriesFailureHandler

beilCrxmarkets:
## Enhance Flexibility in MaxRetriesExceededHandler Configuration
This pull request introduces modifications to the `MaxRetriesFailureHandler` to enhance the flexibility and action possibilities within the `maxRetriesExceededHandler`. Previously, invoking `executionOperations.stop()` universally, even when a `maxRetriesExceededHandler` was defined, limited the potential actions, such as task rescheduling.
Key Changes:
- The `maxRetriesExceededHandler` can now decide independently whether to remove the task, allowing for more nuanced handling (e.g., rescheduling tasks).
- These modifications represent a breaking change, affecting the behavior of the `MaxRetriesFailureHandler`. With no defined handler, the behavior remains unchanged via a default handler executing `executionOperations.stop()`.
- We considered adding a `RetriesFailureHandler`, which would function similarly to `MaxRetriesFailureHandler` but without automatic invocation of `executionOperations.stop()`. However, to avoid user confusion on choosing between handlers, this was not pursued.

## Fixes
[#538](https://github.com/kagkarlsson/db-scheduler/issues/538)

## Reminders

---
cc @kagkarlsson


db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/FailureHandler.java
```diff
@@ -90,10 +93,9 @@ public void onFailure(
       int totalNumberOfFailures = consecutiveFailures + 1;
       if (totalNumberOfFailures > maxRetries) {
         LOG.error(
-            "Execution has failed {} times for task instance {}. Cancelling execution.",
+            "Execution has failed {} times for task instance {}. Invoking maxRetriesExceededHandler",
             totalNumberOfFailures,
             executionComplete.getExecution().taskInstance);
-        executionOperations.stop();
         maxRetriesExceededHandler.accept(executionComplete, executionOperations);
```
kagkarlsson, line 99: I am unsure about the original intent of the `maxRetriesExceededHandler`. The handler should not have taken `executionOperations` as a parameter since the execution had already been stopped.

kagkarlsson:
What is your use-case if I might ask? (what operation rather than `stop()` are you considering)

beilCrxmarkets:
> What is your use-case if I might ask? (what operation rather than `stop()` are you considering)

If the executionOperation is not stoped jet, we could pass `new FailureHandler.OnFailureReschedule<Void>(schedule)::onFailure` or `new FailureHandler.OnFailureRetryLater<Void>(schedule)::onFailure` for the maxRetriesExceededHandler. 

Our RecurringTask is triggered once a day. If this fails, we want to repeat it. If we exceed the maxRetires, we still want the recurring task to be triggered the next day.

kagkarlsson:
Not sure I understand. Why are you using `MaxRetriesFailureHandler` if you do not want the "max"-behavior?

You can always implement your own fully custom failurehandler..

beilCrxmarkets:
> Not sure I understand. Why are you using `MaxRetriesFailureHandler` if you do not want the "max"-behavior?
> 
> You can always implement your own fully custom failurehandler..

At the moment we use our own implemented FailureHandler to get the desired behavior. 

However, I assume that we are not the only ones who need the described behavior, which is why it makes sense to store it in a central location. 

I agree with you that the name “MaxRetriesFailureHandler” implies that after the maximum number of attempts is reached, the instance is terminated. We have therefore also considered proposing to add an additional FailureHandler with the name “RetriesFailureHandler”. However, the existence of a “MaxRetriesFailureHandler” and a “RetriesFailureHandler” can be confusing for beginners at first glance, as the difference is not 100% obvious. 

If we were to start this project from scratch, it would make sense to add only the “RetriesFailureHandler” instead of the “MaxRetriesFailureHandler”, as the latter offers more flexibility. Since the library and thus the “MaxRetriesFailureHandler” have been around for a while and are used by several users as they are, the compromise would be to adapt the “MaxRetriesFailureHandler” accordingly so that it offers more flexibility.

kagkarlsson:
Are you looking for behavior like: 
* up until X retries, use this FailureHandler
* after X retries, use this other FailureHandler
?

beilCrxmarkets:
> Are you looking for behavior like:
> 
> * up until X retries, use this FailureHandler
> * after X retries, use this other FailureHandler
>   ?

Yes, our task will be triggered once a day. If this fails, we want to try again 5 times shortly after (RetriesFailureHandler). After all 5 attempts are used up, we want to try again tomorrow at the usual time (OnFailureReschedule). 

However, with the current implementation of the MaxRetriesFailureHandler, we cannot do this because the task is deleted from the DB after the 5 attempts are exhausted. 

We also combine this with the ExponentialBackoffFailureHandler to get a delay between the 5 additional attempts. This does not cause any problems.

beilCrxmarkets:
@kagkarlsson How should we continue with this topic?

As stated in https://github.com/kagkarlsson/db-scheduler/issues/538#issuecomment-2539864081 @adamalexandru4 has the same issue.

kagkarlsson:
I don't want to add this functionality in to MaxRetries... 

And it is not meant to handle a secondary FailureHandler, the signature is:

```java
    public MaxRetriesFailureHandler(
        int maxRetries,
        FailureHandler<T> failureHandler,
        BiConsumer<ExecutionComplete, ExecutionOperations<T>> maxRetriesExceededHandler) {
      this.maxRetries = maxRetries;
      this.failureHandler = failureHandler;
      this.maxRetriesExceededHandler = maxRetriesExceededHandler;
    }
```

i.e. `BiConsumer<ExecutionComplete, ExecutionOperations<T>>`

(It might be there is a better name for `MaxRetries...`, not sure but since you both made that assumption...)

Your requirement I think should be captured in its own handler. If you want to supply a standard one for the project that is fine 👍. But what should it be called 😅.  `CompositeFailureHandler.primary(handler1).afterTries(5, handler2)` (.afterTime(Duration.ofHours(1), handler3))` (just toying with ideas)

JBabinskas:
Hey,

@kagkarlsson cool library, using it instead of good old Quartz and loving it!

On topic, I think this current `MaxRetriesFailureHandler` implementation is indeed confusing because it not only confused me, but I went around and asked a few colleagues in the office what would happen if one tried to pass any `FailureHandler` as argument for `maxRetriesExceededHandler`. Well, no one said that it would end in an error because `MaxRetriesFailureHandler` always executes `executionOperations.stop();`. I think this is limiting the `MaxRetriesFailureHandler` flexibility for no reason. It is extremely common case to retry some nightly job for X amount of times at night, then give up, and just re-schedule it for next night. And `MaxRetriesFailureHandler` would fit perfectly for that, but I also recognize that changing this behavior would basically be a breaking change, so not sure what would be the best approach here. Currently for those retry than re-schedule cases we just have a custom `FailureHandler` where I basically copy pasted `MaxRetriesFailureHandler` and removed the forced `executionOperations.stop();` 😅 I think part of confusion comes from that one might look at `MaxRetriesFailureHandler` and have an idea that it is a terminating handler, but for most people it just means "retry X amount of times, if all failed, then do X" especially when it takes in something as `MaxRetriesFailureHandler`. Even if it is a breaking change I would vote that `MaxRetriesFailureHandler` would be changed, giving more freedom for the user to decide what to do, including terminating the job completely.

beilCrxmarkets:
@kagkarlsson I agree with @JBabinskas. 

We should adjust the `MaxRetriesFailureHandler`. There are multiple ways. How should we proceed?

omarfi:
We have the exact same use case. 

I also solved it by using a custom "max" retries handler:

```
@Override
    public FailureHandler<OrderFulfilmentTaskData> getFailureHandler() {
            if (executionComplete.getExecution().consecutiveFailures < maxRetries) {
                new OnFailureRetryLater<OrderFulfilmentTaskData>(Duration.ofSeconds(fixedDelayMinutes)).onFailure(executionComplete, executionOperations);
            } else {
                executionOperations.removeAndScheduleNew(TASK_DESCRIPTOR
                        .instance(executionComplete.getExecution().taskInstance.getId())
                        .data((OrderFulfilmentTaskData) executionComplete.getExecution().taskInstance.getData())
                        .scheduledTo(Instant.now().plus(Duration.ofSeconds(24))));
            }
        };
    }
```

I'm also hoping that the library can be shipped with a "CompositeFailureHandler" as proposed in the thread which would be composed of RetryLater and Reschedule- strategy).

Nice library btw.

tstavinoha:
My team came to a similar conclusion as some commenters in this thread.
MaxRetries handler stops/removes the execution, making it impossible to reschedule and thus handle this scenario.

Already in this thread we can find different workarounds being used, with using modulos, rescheduling, etc., and there could be other dirtier hacks in the wild - for what is actually a fairly common use case.

While we can and do implement a custom failure handler, having first-class support for this scenario without needing to reinvent the wheel and work around the internal details of the library would be a big improvement.

beilCrxmarkets:
Hi @kagkarlsson,
I hope you’re doing well and that things have been going smoothly for you.
I opened this PR back in November 2024 to address the issue which we face with the FailureHandler and there’s been good community interest in getting this resolved — several users have commented since then that they’re facing the same issue.
I totally understand that maintaining an open-source project takes a lot of time and effort, often alongside other commitments.
Would you be able to share your thoughts on the current state of this PR? Specifically:
- Any feedback or changes you’d like to see before it could be merged?
- Whether this fits into your roadmap for the project?
- If there’s a different approach you’d prefer for solving this issue?

Even a brief outline of next steps would be super helpful for us to continue collaboratively moving forward.
Thanks a lot for all your work maintaining db-scheduler — it’s a great library and much appreciated by the community!

kagkarlsson:
Sorry for not following up here. I recently put some effort into #791, and I found that there was also a need for a more customizable FailureHandler, so I added some of the ideas from this PR and these discussions to that. I hope that will address some of the issued raised in discussions here. I think there are so many variants that a builder-style will work best, I might consider deprecating the old handlers, at least the ones with confusing signature.

kagkarlsson:
I have migrated most of these changes over to a new PR: #856
A bit cautions to adding more than necessary, there are many concerns here, and all will not be addressed in new PR. Stakeholders are encouraged to provide feedback on the new PR.

kagkarlsson:
I am closing this issue as most of the suggestions here was moved into PR #856

PR #570, opened 2024-12-12
fix: Adjusting primary-key order for SQL Server to reduce deadlocking

### LINKED ISSUE #527: Table schema primary key
Some input from a DBA (not me) regarding the table schema (mssql) that could perhaps be of interest.

**Recommendation based on table analysis**

Based on an analysis of the cardinality of the table columns and index definitions for the table dbo.scheduled_tasks I can see the following from the DBA perspective: 

* The following columns are included as key columns in the PRIMARY KEY (clustered index) of the table:
**task_name**
**task_instance**
* Cardinality analysis of these columns shows that:
**task_name** - contains only two unique values
**task_instance** - contains one unique value per row
* As the order of key columns matters, where you want columns with the highest cardinality first, the order of the columns in the PRIMARY KEY definition for the table should be changed so that task_instance comes first.
* If the order of the columns is changed, the database engine can find the correct row in the table faster and with fewer locks when updating specific rows, as happens in the two UPDATE queries that cause deadlocks. Fewer deadlocks and faster access to the right row lead to less risk of deadlocks.

kagkarlsson:
Adjusting primary-key order for SQL Server to reduce deadlocking after analysis and testing presented in #527

## Fixes

#527

## Reminders

### DIFF
--- a/db-scheduler/src/test/resources/mssql_tables.sql
+++ b/db-scheduler/src/test/resources/mssql_tables.sql
@@ -12,7 +12,7 @@ create table scheduled_tasks
   last_heartbeat       datetimeoffset,
   [version]            bigint         not null,
   priority             smallint,
-  primary key (task_name, task_instance),
+  primary key (task_instance, task_name),
   index execution_time_idx (execution_time),
   index last_heartbeat_idx (last_heartbeat),
   index priority_execution_time_idx (priority desc, execution_time asc)


kagkarlsson:
🎉 This issue has been resolved in `v15.1.1` ([Release Notes](https://github.com/kagkarlsson/db-scheduler/releases/tag/v15.1.1))

PR #594, opened 2025-01-04
#515: improve unresolved task deletion job

### LINKED ISSUE #515: Unresolved execution detection never deletes old tasks if service restarts too often
## Expected Behavior

The unresolved execution detection should delete unresolved tasks once they have been *due* for the duration configured in `deleteUnresolvedAfter`. (Default: 14 days)

## Current Behavior

The unresolved execution detection only deletes unresolved tasks after the *local scheduler instance* has detected the unresolved task for the configured duration. If the service never runs for that duration (due to updates or restarts), unresolved tasks are never deleted.

If this is the intended behaviour, it should be documented.

Workaround: I've reduced `deleteUnresolvedAfter` to 3 days, since the service regularly receives updates and never runs for 14 days straight.

## Steps to Reproduce the bug
1. Create a recurring task
2. Delete the task implementation
3. Restart the scheduler
4. Restart the scheduler before the timeout configured in `deleteUnresolvedAfter`
5. The task won't be deleted until `deleteUnresolvedAfter` after the latest restart

## Context
- DB-Scheduler Version : 14.0.2
- Java Version : 21
- Database and Version : MariaDB 10.11

kamko:
Calculates age of unresolved execution based on it's execution time instead of first scheduler run. 
More in issue: https://github.com/kagkarlsson/db-scheduler/issues/515#issuecomment-2569905886

## Fixes
#515

## Reminders
- [/] Added/ran automated tests
- [/] Update README and/or examples
- [/] Ran `mvn spotless:apply` - also added `.gitattributes` because spotless went mad with CRLF on Windows

---
cc @kagkarlsson

### DIFF
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
@@ -17,6 +17,7 @@ import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.stream.Collectors.toList;
 
 import com.github.kagkarlsson.scheduler.SchedulerState.SettableSchedulerState;
+import com.github.kagkarlsson.scheduler.TaskResolver.UnresolvedTask;
 import com.github.kagkarlsson.scheduler.event.ExecutionInterceptor;
 import com.github.kagkarlsson.scheduler.event.SchedulerListener;
 import com.github.kagkarlsson.scheduler.event.SchedulerListener.SchedulerEventType;
@@ -34,7 +35,9 @@ import com.github.kagkarlsson.scheduler.task.TaskInstanceId;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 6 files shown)


kamko:
Seems weird that the build failed on
````
 Error:  Errors: 
Error:    MssqlClusterTest.test_concurrency_optimistic_locking:68->lambda$test_concurrency_optimistic_locking$1:72 » TaskInstance Failed to add new execution. (task name: Custom, instance id: 2107)
````

**Comment by kagkarlsson** [comment]
Yeah, probably a test-issue

**Comment by kamko** [comment]
Anything I can do with this?

File: db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java [hunk 1]
```diff
@@ -354,8 +357,31 @@ public List<CurrentlyExecuting> getCurrentlyExecutingWithStaleHeartbeat() {
...
+
+    Map<String, Instant> taskToNewestExecutionTime = new HashMap<>();
+    schedulerTaskRepository.getScheduledExecutions(
+        ScheduledExecutionsFilter.all(),
+        execution ->
+            taskToNewestExecutionTime.merge(
+                execution.taskInstance.getTaskName(),
+                execution.executionTime,
+                (oldValue, newValue) -> oldValue.isAfter(newValue) ? oldValue : newValue));
+
```
kagkarlsson, line 369: I would rather avoid this...

Is there perhaps a way to have the TaskResolver also get information on how long the task-name has been unresolved 🤔 (and use that, rather than fetching all executions into mem and checking)

kagkarlsson: I.e. do something [here](https://github.com/kagkarlsson/db-scheduler/blob/master/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java#L792):

Using the execution-time (`Instant executionTime = jdbcCustomization.getInstant(rs, "execution_time");`)

kamko:
Yes, this one is kinda important for some use cases ("versioned jobs"). Our adoption of db-scheduler somehow slowed down so I honestly forgot about this.

PR #605, opened 2025-01-17
fix: recurring tasks get correct default priority without data (#604)

### LINKED ISSUE #604: Recurring tasks without data get priority MEDIUM instead of HIGH
When a recurring task (without data / state) is scheduled, it gets priority MEDIUM (50), while if it has data / state, it gets priority HIGH (90), which matches the default priority for recurring tasks.

This seems like an oversight; recurring tasks should always get their default priority HIGH, no?

geirsagberg:

Include the default priority of a task class when creating an instance without data

## Fixes
#604 

## Reminders

---
cc @kagkarlsson


kagkarlsson (approved):

kagkarlsson:
🎉 This issue has been resolved in `v15.1.2` ([Release Notes](https://github.com/kagkarlsson/db-scheduler/releases/tag/v15.1.2))

PR #606, opened 2025-01-17
feat: set default priority

geirsagberg:

This PR makes it possible to set a default priority for all built-in task types.

This makes it possible to e.g. define recurring tasks that only will run when the queue is otherwise empty.

## Reminders

---
cc @kagkarlsson

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,14 @@ Set the priority per instance using the `TaskInstance.Builder`:
             .scheduledTo(Instant.now()));
 ```
 
+You can also set the default priority for all tasks of a given type:
+
+```java
+Tasks.recurring("my-task", FixedDelay.ofSeconds(5))
+    .defaultPriority(Priority.LOW)
+    .execute(...);
+```
+
 **Note:**
 * When enabling this feature, make sure you have the new necessary indexes defined. If you
 regularly have a state with large amounts of executions both due and future, it might be beneficial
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -250,6 +250,14 @@ Set the priority per instance using the `TaskInstance.Builder`:
             .scheduledTo(Instant.now()));
 ```
 
+You can also set the default priority for all tasks of a given type:
+
+```java
+Tasks.recurring("my-task", FixedDelay.ofSeconds(5))
+    .setDefaultPriority(Priority.LOW)
```
kagkarlsson, line 257: Maybe skip set, i.e. `.defaultPriority(Priority.LOW)`, to harmonize with the other methods..

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/helper/CustomTask.java
```diff
@@ -36,7 +36,25 @@ public CustomTask(
...
       DeadExecutionHandler<T> deadExecutionHandler) {
-    super(name, dataClass, failureHandler, deadExecutionHandler);
+    this(
+        name,
+        dataClass,
+        scheduleOnStartup,
+        defaultExecutionTime,
+        failureHandler,
+        deadExecutionHandler,
+        DEFAULT_PRIORITY);
```
kagkarlsson, line 46: Should this be a field on this class as well? 🤔

kagkarlsson (commented):
Looks good 👍, just a couple of small comments.

(Also, this will not work when the client is using a `TaskDescriptor` to schedule new executions, i.e. without a reference to the `Task`, but I suppose the most important fix here is for recurring tasks)

README.md
geirsagberg: Problem was that we use `defaultPriority` for the name of the field, so the method can't have the same name.

Can rename to `_defaultPriority`, but no other private vars have underscores 😅

geirsagberg: Will fix it by using `priority` for the private field.

geirsagberg: ah nvm, seems its ok to have same name of private field and method

geirsagberg:
> Looks good 👍, just a couple of small comments.
> 
> (Also, this will not work when the client is using a `TaskDescriptor` to schedule new executions, i.e. without a reference to the `Task`, but I suppose the most important fix here is for recurring tasks)

You can set the priority on either the `TaskInstance.Builder`, or on the instance itself, so should cover most use cases I believe

PR #628, opened 2025-02-07
feat: Extend DBSchedulerCustomizer with datasource method

### LINKED ISSUE #626: Autoconfiguration: Add an option to specify which DataSource should be used.
Currently, the `DbSchedulerAutoConfiguration` selects the primary `DataSource` bean by default, which may not always be the desired behavior.

Although it’s possible to override the default `Scheduler` bean, this approach can be somewhat cumbersome.

An alternative solution could be to either update the `DbSchedulerCustomizer` to include a datasource method or implement a custom `DataSourceSupplier`. 

What are your thoughts on these approaches?

Two different implementations; 
https://github.com/kagkarlsson/db-scheduler/pull/627
https://github.com/kagkarlsson/db-scheduler/pull/628

NicklasWallgren:
Fixes #626 

## Reminders

---
cc @kagkarlsson


kagkarlsson:
I think this one is better

kagkarlsson:
🎉 This issue has been resolved in `v15.3.0` ([Release Notes](https://github.com/kagkarlsson/db-scheduler/releases/tag/v15.3.0))

PR #642, opened 2025-03-07
Update README.md

Hronom:
Fix `getId()` in example

## Reminders
- [ ] Ran `mvn spotless:apply`

---
cc @kagkarlsson


kagkarlsson:
Not sure if this was a bug? Isnt't `.id` referring to the field in `MyTaskData`?

Hronom:
@kagkarlsson for a moment to me it appears as a wrong example, as attempt to print id of task. Maybe this is what confused e.g. we have id of task and we have extra id in `MyTaskData`

PR #670, opened 2025-04-16
fix: Reduce log-level to DEBUG when missing a single heartbeat update

### LINKED ISSUE #654: dbscheduler 15.1.1 WARN-logging in c.g.k.s.jdbc.JdbcTaskRepository  - Did not update heartbeat
[db-scheduler-housekeeper-pool-5-thread-1] WARN  c.g.k.s.jdbc.JdbcTaskRepository  - Did not update heartbeat. Execution must have been removed or rescheduled(i.e. CompletionHandler ran and finished just before heartbeat-update). This is a race-condition that may occur, but is very unlikely.

We have a few of these error-messages a day. What are we supposed to do with them, or how can we prevent them?

Doesn' t feel right to just teach our ops-people to ignore these WARN-messages.

kagkarlsson:
See issue #654. This fix reduces that logging from WARN to DEBUG. If there are consecutive failures for heartbeats, there will still be a WARN (handled on the outside)

## Fixes
Fixes #654

## Reminders
- [ ] Ran `mvn spotless:apply`

### DIFF
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
@@ -647,7 +647,7 @@ public class JdbcTaskRepository implements TaskRepository {
     if (updated == 0) {
       // There is a race-condition: Executions are not removed from currently-executing until after
       // the execution has been updated in the database, so this might happen.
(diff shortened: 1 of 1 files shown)


kagkarlsson:
🎉 This issue has been resolved in `v15.4.2` ([Release Notes](https://github.com/kagkarlsson/db-scheduler/releases/tag/v15.4.2))

Issue #666, opened 2025-04-15
Move micro-jdbc into db-scheduler for simpler maintenance and PRs


kagkarlsson:
Maintaining micro-jdbc in its own dependency adds to the complexity of making changes on the level of jdbc-access. Suggest moving it into db-scheduler.

kagkarlsson:
Fixed in a previous release

PR #682, opened 2025-04-23
feat: Inline micro jdbc

kagkarlsson:
Rather than maintaining it as a separate project, inline it to simplify both maintenance and understanding.


kagkarlsson:
Fixes #666

PR #701, opened 2025-05-27
fix: Use AutodetectJdbcCustomization by default in TestHelper (#640)

### LINKED ISSUE #640: `TestHelper.createManualScheduler()` does not work for MySQL 8
Running a following test
```
package org.example;

import com.github.kagkarlsson.scheduler.task.helper.RecurringTask;
import com.github.kagkarlsson.scheduler.task.helper.Tasks;
import com.github.kagkarlsson.scheduler.task.schedule.FixedDelay;
import com.github.kagkarlsson.scheduler.testhelper.ManualScheduler;
import com.github.kagkarlsson.scheduler.testhelper.SettableClock;
import com.github.kagkarlsson.scheduler.testhelper.TestHelper;
import com.zaxxer.hikari.HikariConfig;
import com.zaxxer.hikari.HikariDataSource;

import javax.sql.DataSource;
import java.time.Duration;
import java.util.List;

public class Main {
    public static void main(String[] args) {
        DataSource dataSource = dataSource();
        RecurringTask<Void> task = Tasks
                .recurring("test-task", FixedDelay.ofSeconds(10))
                .execute((taskInstance, executionContext) -> System.out.println("Hello, world!"));
        SettableClock clock = new SettableClock();

        ManualScheduler scheduler = TestHelper
                .createManualScheduler(dataSource, List.of(task))
                .clock(clock)
                .build();

        scheduler.start();
        scheduler.tick(Duration.ofSeconds(20));
        scheduler.runAnyDueExecutions();
    }

    private static DataSource dataSource() {
        HikariConfig config = new HikariConfig();
        config.setUsername("mysql");
        config.setPassword("mysql");
        config.setJdbcUrl("jdbc:mysql://localhost:3306/test");
        return new HikariDataSource(config);
    }
}
```
against MySQL 8 results in a failure `Caused by: java.sql.SQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'OFFSET 0 ROWS FETCH FIRST 30 ROWS ONLY' at line 1`.

The reason for that is that `TestHelper` will always use `DefaultJdbcCustomization`:
- https://github.com/kagkarlsson/db-scheduler/blob/master/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/testhelper/TestHelper.java#L79
- https://github.com/kagkarlsson/db-scheduler/blob/master/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/testhelper/TestHelper.java#L90

and `DefaultJdbcCustomization` tries to use ANSI pagination constructs that aren't supported in MySQL.

It would be better to honour `jdbcCustomization` set on the builder and fall back to `AutodetectJdbcCustomization` if not set. It would boil down to something like

```
      ...
      final JdbcCustomization jdbcCustomization =
          ofNullable(this.jdbcCustomization)
              .orElseGet(
                  () -> new AutodetectJdbcCustomization(dataSource, alwaysPersistTimestampInUTC));
      final JdbcTaskRepository schedulerTaskRepository =
          new JdbcTaskRepository(
              dataSource,
              true,
              jdbcCustomization,
              tableName,
              taskResolver,
              new SchedulerName.Fixed("manual"),
              serializer,
              enablePriority,
              clock);
      final JdbcTaskRepository clientTaskRepository =
          new JdbcTaskRepository(
              dataSource,
              commitWhenAutocommitDisabled,
              jdbcCustomization,
              tableName,
              taskResolver,
              new SchedulerName.Fixed("manual"),
              serializer,
              enablePriority,
              clock);
      ...
```

in `ManualSchedulerBuilder.build()`.

I run all tests with this small change and they are all green. @kagkarlsson I can contribute a PR if you are interested. It would be even better to have some (integration) test(s) but that would I think require adding support for running an embedded MySQL instance (similar to what I can see for PostgreSQL).

hipnoizz:
Previously used DefaultJdbcCustomization would try to apply ANSI syntax for paging, which is not supported by MySQL and MariaDB databases, and would result in exceptions when trying to use TestHelper with those databases.

## Fixes
#640

## Reminders

---
cc @kagkarlsson

### DIFF
--- a/db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/compatibility/CompatibilityTest.java
+++ b/db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/compatibility/CompatibilityTest.java
@@ -30,7 +30,6 @@ import com.github.kagkarlsson.scheduler.helper.TestableListener;
 import com.github.kagkarlsson.scheduler.jdbc.AutodetectJdbcCustomization;
 import com.github.kagkarlsson.scheduler.jdbc.JdbcCustomization;
 import com.github.kagkarlsson.scheduler.jdbc.JdbcTaskRepository;
-import com.github.kagkarlsson.scheduler.jdbc.PostgreSqlJdbcCustomization;
 import com.github.kagkarlsson.scheduler.task.Execution;
 import com.github.kagkarlsson.scheduler.task.SchedulableInstance;
 import com.github.kagkarlsson.scheduler.task.SchedulableTaskInstance;
@@ -454,7 +453,7 @@ public abstract class CompatibilityTest {
   }
 
   @Test
-  void test_compatibility_test_helper() {
+  void test_compatibility_manual_scheduler() {
     final SettableClock clock = new SettableClock();
     final ManualScheduler scheduler =
         (ManualScheduler)


db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/compatibility/CompatibilityTest.java
```diff
@@ -449,6 +453,22 @@ public void test_jdbc_customization_supports_timestamps_in_utc() {
     assertThat(taskRepo.getExecution(instance1).get().executionTime, is(zonedDateTime.toInstant()));
   }
 
+  @Test
+  void test_compatibility_test_helper() {
```
kagkarlsson, line 457: Maybe `test_compatibility_manual_scheduler`

PR #704, opened 2025-06-09
feat: add priority pools to prevent high-priority task starvation (#702)

### LINKED ISSUE #702: Feature: Fast lane / priority workers
We have some tasks that are higher priority than others, and several of our tasks are long-running (10-30 minutes). Currently, if a high-priority task is queued and the workers are busy with low-priority tasks, it will have to wait for the next available one.

It would be very useful to assign separate "high priority" worker threads, that ignore any tasks below a given priority!

geirsagberg:
Allows high-priority tasks to execute immediately even when all default threads are busy with lower-priority work. 

As described in #702 we had a need for "fast lane" pools for high priority tasks, so I spent the weekend working with Claude and ended up with a solution that uses multiple executors with separate thread pools. Open for feedback and alternative solutions!

Also, I added guidelines for Claude and also Jetbrains Junie.

## Fixes
#702

## Reminders
- [X] Update README and/or examples

---
cc @kagkarlsson

### DIFF
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
@@ -40,6 +40,7 @@ import java.util.Optional;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 import javax.sql.DataSource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/LockAndFetchCandidates.java
```diff
@@ -72,11 +76,48 @@ public LockAndFetchCandidates(
     upperLimit = pollingStrategyConfig.getUpperLimit(threadpoolSize);
   }
 
+  public LockAndFetchCandidates(
+      List<Executor> executors,
```
geirsagberg, line 80: Same here, could potentially deprecate the old one

PR #720, opened 2025-07-15
feat: Expose Clock from builder and implement in Spring Boot AutoConfiguration

deximat:
- Added clock(Clock) method to SchedulerBuilder so users can provide their own clock for Scheduler
- In order not to have to think about Waiter caching old Clock I moved Waiter creation to build method, so it always uses latest Clock from builder field
- Had to build Waiter manually in TestHelper for ManualScheduler.

Note: This changes current behavior, but I think it's a bug. 
Currently Waiter in ManualScheduler is using SystemClock instead of newly set clock, I changed that because Waiter is now created in the end so it will always use up-to date Clock instance, in case this is desired behavior for some reason, let me know and I will fix it. 

## Reminders

---
cc @kagkarlsson

### DIFF
--- a/db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/autoconfigure/DbSchedulerAutoConfiguration.java
+++ b/db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/autoconfigure/DbSchedulerAutoConfiguration.java
@@ -13,10 +13,12 @@
  */
 package com.github.kagkarlsson.scheduler.boot.autoconfigure;
 
+import com.github.kagkarlsson.scheduler.Clock;
 import com.github.kagkarlsson.scheduler.PollingStrategyConfig;
 import com.github.kagkarlsson.scheduler.Scheduler;
 import com.github.kagkarlsson.scheduler.SchedulerBuilder;
 import com.github.kagkarlsson.scheduler.SchedulerName;
+import com.github.kagkarlsson.scheduler.SystemClock;
 import com.github.kagkarlsson.scheduler.boot.config.DbSchedulerCustomizer;
 import com.github.kagkarlsson.scheduler.boot.config.DbSchedulerProperties;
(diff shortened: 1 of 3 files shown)


deximat:
Now user can just define Clock bean and it will automatically be used by AutoConfiguration.

kagkarlsson:
Thank you for contributing!

What was the use-case here? Override clock used in tests that uses Spring to instantiate the Scheduler?

osoykan:
It is a good idea to override Clock interface from outside, I agree. It provides an ability to alter time while writing e2e and component tests.

kagkarlsson:
🎉 This issue has been resolved in `v16.2.0` ([Release Notes](https://github.com/kagkarlsson/db-scheduler/releases/tag/v16.2.0))

deximat:
Thanks for fast response!

MKorsch:
Unfortunately, this change breaks our build. We have a custom Spring Boot configuration that exposes a `java.time.Clock` bean:

```kotlin
import org.springframework.context.annotation.Bean
import org.springframework.context.annotation.Configuration
import java.time.Clock

@Configuration
class ClockConfiguration {

    @Bean
    fun clock(): Clock = Clock.systemUTC()
}
```

When starting the service, we encounter a `BeanDefinitionOverrideException`. The `@ConditionalOnMissingBean` in `DbSchedulerAutoConfiguration` is not effective in this case because the two beans have different types: ours is `java.time.Clock`, while the auto-configured bean is `com.github.kagkarlsson.scheduler.Clock`.

This leads to a name collision (clock) despite the type difference.

Is there a recommended way to resolve this without having to use a specific `@Qualifier` for the Clock bean throughout our code?

Would it be feasible for db-scheduler to rename its bean to something more specific, e.g., `dbSchedulerClock`, to avoid conflicts with common types like `java.time.Clock`?

deximat:
I agree. It should be fine to be renamed to dbSchedulerClock, so it more fully qualifies bean name, since it it is not standard clock, but db scheduler's.

kagkarlsson:
Would be much appreciated! 😊

PR #765, opened 2025-11-06
refactor: extract spring common module

jlafourc:
## Summary
- add `db-scheduler-spring-common` to host the shared Spring configuration/startup classes
- keep `db-scheduler-spring-boot-starter` intact for existing consumers by simply wiring it to the new shared module
- touch only the minimum POMs and file moves needed so the SB3 starter diff stays small ahead of the SB4 follow-up

## Stability
- preserves the existing starter artifact id and package structure, so downstream builds and autoconfiguration metadata remain unchanged

## Reminders

---
cc @kagkarlsson

### DIFF
--- a/db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/autoconfigure/DbSchedulerAutoConfiguration.java
+++ b/db-scheduler-boot-starter/src/main/java/com/github/kagkarlsson/scheduler/boot/autoconfigure/DbSchedulerAutoConfiguration.java
@@ -130,5 +130,4 @@ public class DbSchedulerAutoConfiguration {
   static LazyInitializationExcludeFilter eagerDbSchedulerStarter() {
     return LazyInitializationExcludeFilter.forBeanTypes(DbSchedulerStarter.class);
   }
-
 }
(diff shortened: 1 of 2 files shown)


kagkarlsson:
Was it not possible to move the properties as well?

jlafourc:
Yes, is that what you had in mind ?

db-scheduler-spring-common/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>com.github.kagkarlsson</groupId>
+            <artifactId>db-scheduler</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-autoconfigure</artifactId>
+        </dependency>
```
kagkarlsson: This module should probably not contain a boot-dep. Is that because of the DbSchedulerProperties? (is there anything there that breaks?

kagkarlsson:
> You consider drop compatibility with boot 3 in the next major version of db-scheduler ?

It will need to live a bit longer than that most likely. At some point it will be dropped, but I would say at least 1-2 years of boot3 support.

db-scheduler-spring-common/pom.xml
kagkarlsson: Alright, lets try that

jlafourc: OK should be better now

PR #779, opened 2026-01-16
fix: Add default value for execution time column

lhoegh:
Add default value for execution time column, to prevent mysql/mariadb from automatically adding ON UPDATE CURRENT_TIMESTAMP. This strange quirk is explained, for example, in this stackoverflow: https://stackoverflow.com/questions/34355965/mysql-timestamp-column-is-auto-updated-why

I'm not sure if the auto updating of the execution time breaks anything, but it definitely is confusing behavior when using mysql/mariadb.

I  don't think running the tests etc. really does anything in this case.

---
cc @kagkarlsson

### DIFF
--- a/db-scheduler/src/test/resources/mariadb_tables.sql
+++ b/db-scheduler/src/test/resources/mariadb_tables.sql
@@ -3,7 +3,7 @@ create table test.scheduled_tasks (
   task_name varchar(100) not null,
   task_instance varchar(100) not null,
   task_data blob,
-  execution_time timestamp(6) not null,
+  execution_time timestamp(6) not null default current_timestamp(6),
   picked BOOLEAN not null,
   picked_by varchar(50),
   last_success timestamp(6) null,
--- a/db-scheduler/src/test/resources/mysql_tables.sql
+++ b/db-scheduler/src/test/resources/mysql_tables.sql
@@ -2,7 +2,7 @@ create table test.scheduled_tasks (
   task_name varchar(100) not null,
   task_instance varchar(100) not null,
   task_data blob,
-  execution_time timestamp(6) not null,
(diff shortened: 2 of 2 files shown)


db-scheduler/src/test/resources/mariadb_tables.sql
```diff
@@ -3,7 +3,7 @@ create table test.scheduled_tasks (
   task_name varchar(100) not null,
   task_instance varchar(100) not null,
   task_data blob,
-  execution_time timestamp(6) not null,
+  execution_time timestamp(6) not null default current_timestamp(6),
```
kagkarlsson: Interesting.
How about `not null default null`?

lhoegh: Sorry, I didn't mean to be impolite, the library looks super helpful - thanks for the effort you have put into it.

After creating the table with the script provided, I noticed that our MariaDB had added `default current_timestamp(6) on update current_timestamp(6)` to the table definition. This makes the db autoincrement the execution_time every time the row is updated. I looked at the equivalent table definitions for the other databases, and it didn't appear to be intended behavior.

I do not know whether this would cause issues when running db-scheduler in prod - I only decided to try adding it to our codebase yesterday, so its quite early days for me still. Either way i have made the above change in our migrations, so its not a change I need - I just figured, given the comment at the top of mariadb_tables.sql, that you might be interested in a quick pr :)

kagkarlsson: You also agree that `DATETIME` is a better choice?

kagkarlsson:
Closing since PR change type to `DATETIME` has been merged.

Issue #775, opened 2025-12-09
StandardSchedulerClient.reschedule may throw ExecutionException when multiple threads/nodes are rescheduling the same task instance.


MillQK:
Currently, we have the next case in our application - we have some background task which maintain some state and on some events we would like to reduce latency and run the task as soon as possible and for that we use `schedule` function with `ScheduleOptions.WHEN_EXISTS_RESCHEDULE` option. Couple of times we faced `TaskInstanceCurrentlyExecutingException`, but in our case this's fine and we can just skip the exception, but `ExecutionException` is a little bit different, because it states in the message: `Indicates a bug.`. In my opinion it's completely valid, because in `StandardSchedulerClient.reschedule` call `taskRepository.getExecution` doesn't lock the execution in JdbcTaskRepository, so 2+ threads/nodes can easily pick the same task instance with the same version and then try to reschedule it, but only one succeed, because in `JdbcTaskRepository.rescheduleInternal` there's a version check in `where` clause.

Are we doing something wrong or this case is definitely valid and possible and doesn't indicate a bug?

kagkarlsson:
SchedulerClient.reschedule should not enforce version-check and not throw the "indicates a bug"-exception, this was originally for internal rescheduling that happens when task is complete and thus in the context of a lock. SchedulerClient.reschedule need to use a separate implementation of reschedule that does not fail when concurrent update

Issue #315, opened 2022-09-05
On complete, disable a task for ever, even across db-scheduler restarts


reda-alaoui:
Hello,

Sometimes we want to create tasks that should execute a limited number of times. Often these tasks should execute only once.
They are often data manipulation tasks that should run in the background without disturbing the application nominal execution.

For now, we found a workaround that seems to work:
```java
Tasks.custom("foo", Void.class)
        .scheduleOnStartup("once", null, Function.identity())
        .execute(
            (taskInstance, executionContext) -> {
              LOG.info("Hello World");
              return (executionComplete, executionOperations) ->
                  executionOperations.reschedule(
                      executionComplete,
                      executionComplete.getTimeDone().plus(1_000_000, ChronoUnit.DAYS));
            });
```

On complete, we reschedule the task to be executed in `1 000 000 days`.

I find this workaround clumsy. This is a legacy I don't want to transmit to java developper of year 4760 :D
I think this feature should be explicitely part of `db-scheduler`.

I have 2 ideas of implementation about that:
- Make `scheduled_task.execution_time` nullable. If null, we would consider the task as terminated forever.
- Add a boolean column `disabled` to `scheduled_task` to mark the task as disabled forever.

On the java side, we would add a new `CompletionHandler` implementation. Maybe named `OnCompleteDisable`?

What do you think?

reda-alaoui:
> Why not remove it completely (which is supported)?

## If you mean removing it from the `scheduled_tasks` table
According to my tests, on next application reboot (without changing the code declaring the task), the task will be rescheduled and executed.

## If you mean removing it from the java code
We want the task to stay disabled until we remove its declaration from the java code.

reda-alaoui:
So what do you think? :blush:

kagkarlsson:
I think I see your use-case, you would like to use it similar to what you do with database-migrations using liquibase or flyway. Run once and keep the history to ensure it never runs again. This requires history, which I have so far tried to avoid. 
I suppose this might be a useful thing to be able to do, and you would just have to take care not to misuse (i.e. ever growing large table, in which case you are better off storing the history in another table)

I am already considering introducing a `state` column, which would cover you case as well. States might be:

* active
* dead-letter-queue / failed (failed too many times and requires manual intervention)
* paused / disabled.  (make it possible to temporarily pause/disable recurring tasks via UI or similar)
* complete (your case?)

reda-alaoui:
[...]

Issue #27, opened 2018-04-20
Add ability to "pause" or disable tasks (recurring)


kagkarlsson:
Add a method `SchedulerClient.pause(..)`  to enable temporarily pausing the execution, i.e. not run when due.

codewithrajesh:
Hi, @kagkarlsson how is currently a recurring task is stopped?

kagkarlsson:
If you deploy a new version of the code where the `Scheduler` is not started with knowledge of that task-type it will be removed after 14days (default value). (and it will not be executed in the meantime)

Or are you talking of more temporary solutions?

codewithrajesh:
Apologies for the late response.
Actually, I have two questions now.
1. Say, the scheduler is started with task-name & implementation mapping. However, I want to stop the future occurrence of the schedule. how can we achieve that?
2. In another scenario, say, I have scheduled a job(onetime or recurring) and task-name & implementation mapping is also in place(before scheduler start) but after some time the scheduler instance is killed because of some issue, on restart will the scheduler be able to pick the task which was to schedule in the future?

kagkarlsson:
1. If you don't want to redeploy a version of your app where the task is removed (the easy way, as mentioned above), you can for recurring tasks use `Scheduler.cancel(TaskInstanceId.of("your-task-name", "recurring"))`  (for non-recurring, you need to use the correct `instance-id`

2. Yes the future executions are persistent, they live in the database table `scheduled_tasks`. However, if the scheduler is killed (non-gracefully) while it is running an execution, the execution will not be unlocked until it some scheduler-instance consider it to be "dead". You can read about that under [Dead executions](https://github.com/kagkarlsson/db-scheduler#dead-executions). Default behavior is that "dead" executions get rescheduled to a new time, but that can be overridden in the builder.

codewithrajesh:
Thanks for the response.
For us, The task is an I/O bound operation. basically, it does a POST request at a given time on a given endpoint. I don't see that information(the endpoint) stored in the "scheduled_tasks" table.
Maybe we need to make use of _"task_data"_ field?

kagkarlsson:
Yes, if you have some context data at scheduling-time that you want the task to have access to at execution-time you need to either encode it in the instance-id string, or use custom task-data

setraj:
Thanks, i was able to persist the task data.
But still, i am not clear on how on the startup the scheduler will pick up the future tasks. I keep getting this **error**.
[...]

PR #791, opened 2026-02-23
feat: State-support (active, failed, complete, record, paused)

kagkarlsson:
# Deactivation Support / State-support

Add functionality to put executions in a deactivated state, stopping further executions,
but keeping it as a historic record for deduplicatation or possibility to resume.

## High-level

- `state` column for holding the state of the execution.
- Only active executions are picked and run.
- Non-active state = "deactivated"
- CompletionHandlers and FailureHandlers now get the option to deactivate the execution and set to a
  deactivated state (in addition to the previous options: remove and reschedule)

### Schema/query changes

- Add nullable `state` column (text/varchar). (null considered 'active')
- Existing queries for fetching due is updated with state-condition: `(state=ACTIVE or state is null)`

### Execution states

|      State      |                                                                  Description                                                                  |
|-----------------|-----------------------------------------------------------------------------------------------------------------------------------------------|
| (null) / ACTIVE | Active, scheduled for execution (default, backward compatible)                                                                                |
| COMPLETE        | Execution completed successfully and is kept for deduplication. Can be deleted by automatic process after configured time.                    |
| PAUSED          | Execution has been paused, typically via manual action. Will not execute until resumed.                                                       |
| RECORD          | Execution completed and is kept indefinitely as a historic record. Similar to liquibase migration records - ensures something runs only once. |
| FAILED          | Execution is marked as permanently failed by its FailureHandlers. Must be manually triggered to retry.                                        |
| WAITING         | Execution is waiting for activation to run. Semantically different from PAUSED, which is a "manual" state, this is meant to be used by automatic processes.                                                                                                   |

### SchedulerClient

- Let `client.getScheduledExecutions(...)` also return deactivated executions.
  Not by default, but opt-in via filter
- New methods on `ScheduledExecution`:
  - `.getState()`
- New/modified filter methods
  - `.withIncludeDeactivated()`
  - `ScheduledExecutionsFilter.active()`
  - `ScheduledExecutionsFilter.deactivated()`
  - Deprecate `ScheduledExecutionsFilter.all()` (`active` new default)
- Use `client.reactivate(TaskInstanceId, Instant)` to reschedule deactivated executions. Will reset
  lastSuccess, lastFailure and consecutiveFailures
  - RecurringTasks will not have option to reschedule according to Schedule
- Use `client.deactivate(TaskInstanceId, State)` to manually deactivate a scheduled execution. Will
  keep lastSuccess, lastFailure and consecutiveFailures

### Jobs

- Introduce housekeeping-job that regularly deletes deactivated executions. Default 14d for all
  except `State.RECORD` which is kept indefinately.

### CompletionHandlers (changes)

- `.onCompleteDeactivate(State)` - deactivate with given state
- `.onCompleteKeepRecord()` - convenience for `.onCompleteDeactivate(State.RECORD)`

### FailureHandlers (changes)

- Builds on the FailureHandler.maxRetries(...) builder introduced in #856; adds thenDeactivate(...)

### OnStartup

- Recurring tasks that have been deactivated should not be reactivated when scheduler restarts.
  Logs a WARN for now if Recurring is paused.

## Other things

- Only active executions should be considered by dead execution detection
- `state = null` means active/scheduled (existing rows should work without migration)
- `state` column will be required (migration needed for existing schemas)
- Consider adding future column `state_details (TEXT)` / `failure_details (TEXT)` to hold exceptions
  from failure

### Possible minor breaking changes

- reschedule throw `TaskInstanceException` instead of `ExecutionException`

**Disclaimer**: parts of the code in this PR was generated via Claude Code, but the changes have been thoroughly reviewed.

## Fixes

* #27
* #315
* #775

## Reminders


db-scheduler-spring-boot-starter-parent/db-scheduler-spring-boot3-starter/src/test/resources/schema.sql
```diff
@@ -11,5 +11,6 @@ create table if not exists scheduled_tasks (
 	last_heartbeat TIMESTAMP WITH TIME ZONE,
 	version BIGINT,
 	priority INT,
+	state varchar(20),
```
kagkarlsson, line 14: Consider adding `last_failure_text` as well, for future feature. To avoid too frequent schema-changes

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/exceptions/TaskInstanceNotActiveException.java
```diff
@@ -0,0 +1,29 @@
...
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.kagkarlsson.scheduler.exceptions;
+
+import com.github.kagkarlsson.scheduler.task.TaskInstanceId;
+import java.io.Serial;
+
+public class TaskInstanceNotActiveException extends TaskInstanceException {
```
kagkarlsson: `TaskInstanceAlreadyDeactivatedException`? (Check with all uses)

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/exceptions/TaskInstanceNotDeactivatedException.java
```diff
@@ -0,0 +1,29 @@
...
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.kagkarlsson.scheduler.exceptions;
+
+import com.github.kagkarlsson.scheduler.task.TaskInstanceId;
+import java.io.Serial;
+
+public class TaskInstanceNotDeactivatedException extends TaskInstanceException {
```
kagkarlsson: `TaskInstanceAlreadyActiveException`?

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/ExecutionUpdate.java
```diff
@@ -0,0 +1,232 @@
...
+      setValues.add((ps, index) -> ps.setString(index, pickedBy.value));
+    }
+
+    if (lastSuccess != null) {
+      setColumns.add("last_success");
+      setValues.add(
+          (ps, index) -> {
+            jdbcConfig.customization().setInstant(ps, index, lastSuccess.value);
+          });
+    }
```
kagkarlsson: Pattern could be improved.. ideally some sort of query-builder, `.and("last_success", lastSuccess.value)`. Possibly a later refactoring..

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/ExecutionUpdate.java
```diff
@@ -0,0 +1,232 @@
...
+    setValues.add((ps, index) -> ps.setLong(index, versionToUpdate));
+
+    LOG.debug("ExecutionUpdate query: {}", query);
+    int updatedRows = jdbcConfig.runner().execute(query, toPreparedStatementSetter(setValues));
+    if (updatedRows != 1) {
+      throw new TaskInstanceException(
+          "Expected one execution to be updated, but updated " + updatedRows + ". Indicates a bug.",
+          taskInstance.getTaskName(),
+          taskInstance.getId());
+    }
```
kagkarlsson: Not sure if this might be too strict, are we only using update from a "picked"-context? Probably not, and then this might also update 0 rows, in a situation with a concurrent-update. 
Probably need another method, `tryUpdateSingle` that returns a result indicating the result (>1 updated is still an exception)

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
```diff
@@ -206,6 +231,7 @@ private void setInsertParameters(ScheduledTaskInstance value, PreparedStatement
     ps.setString(index++, taskInstance.getId());
     jdbcCustomization.setTaskData(ps, index++, serializer.serialize(taskInstance.getData()));
     jdbcCustomization.setInstant(ps, index++, value.getExecutionTime());
+    ps.setString(index++, State.ACTIVE.name());
```
kagkarlsson, line 236: Limit state to 4 chars?

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
```diff
@@ -329,6 +355,17 @@ public void getScheduledExecutions(
...
+  @Override
+  public void getDeactivatedExecutions(Consumer<Execution> consumer) {
+    QueryBuilder q = QueryBuilder.selectFromTable(tableName);
+    q.andCondition(new DeactivatedCondition());
+
+    jdbcRunner.query(
+        q.getQuery(jdbcCustomization),
+        q.getPreparedStatementSetter(jdbcCustomization),
+        new ExecutionResultSetConsumer(consumer, true, false));
+  }
```
kagkarlsson: If this is not used from main/ code, it should be removed

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
```diff
@@ -474,15 +511,58 @@ public void remove(Execution execution) {
...
+      update.lastSuccess(deactivateUpdate.lastSuccess().value());
+    }
+
+    if (deactivateUpdate.lastFailure() != null) {
+      update.lastFailure(deactivateUpdate.lastFailure().value());
+    }
+
+    if (deactivateUpdate.state() != null) {
+      update.state(deactivateUpdate.state().value());
+    }
```
kagkarlsson: The state to deactivate to should always be set. Throw if not

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/DeactivateUpdate.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.time.Instant;
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+@NullMarked
+public record DeactivateUpdate(
+    @Nullable NewValue<Instant> lastSuccess,
+    @Nullable NewValue<Instant> lastFailure,
+    @Nullable NewValue<State> state) {
```
kagkarlsson: Not nullable, should always be set

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
```diff
@@ -716,6 +774,45 @@ public int removeExecutions(String taskName) {
...
+            + "    ORDER BY execution_time"
+            + jdbcCustomization.getQueryLimitPart(limit)
+            + "  ) sub"
+            + "  WHERE sub.task_name = "
+            + tableName
+            + ".task_name"
+            + "  AND sub.task_instance = "
+            + tableName
+            + ".task_instance"
+            + ")";
```
kagkarlsson: Is there a compatibility-test for this, verifying only old is deleted?

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/helper/ScheduleRecurringOnStartup.java
```diff
@@ -60,8 +61,22 @@ public void apply(SchedulerClient scheduler, Clock clock, Task<T> task) {
     }
 
     if (preexistingExecution.isPresent()) {
+      ScheduledExecution<Object> execution = preexistingExecution.get();
+      var state = preexistingExecution.get().getState();
```
kagkarlsson: `preexistingExecution.get()` -> `execution`

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/helper/Tasks.java
```diff
@@ -296,8 +298,14 @@ public OneTimeTaskBuilder<T> defaultPriority(int priority) {
       return this;
     }
 
+    public OneTimeTaskBuilder<T> onCompleteKeep(State state) {
```
kagkarlsson: Consider `onCompleteDeactivate` or `onCompleteKeepDeactivated`

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/Execution.java
```diff
@@ -63,6 +69,14 @@ public boolean isPicked() {
     return picked;
   }
 
+  public boolean isActive() {
+    return getState() == State.ACTIVE;
+  }
+
+  public State getState() {
+    return state == null ? State.ACTIVE : state;
+  }
```
kagkarlsson, line 78: I think this can be removed now, state is never set as null here I think

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/FailureHandler.java
```diff
@@ -93,14 +106,109 @@ public void onFailure(
...
+    /** After max retries, remove the execution and call the callback. */
+    public FailureHandler<T> thenRemove(MaxRetriesExceededCallback callback) {
+      return then(
+          (complete, ops) -> {
+            LOG.error(
+                "Execution has failed max retries for task instance {}. Removing.",
+                complete.getExecution().taskInstance);
+            ops.remove();
+            callback.maxRetriesExceeded(complete);
+          });
```
kagkarlsson: Double check that we logged error previously as well (might be handled at another level)

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/State.java
```diff
@@ -0,0 +1,47 @@
...
+   * to liquibase migration records - can ensure something runs only once, guaranteed.
+   */
+  RECORD,
+
+  /**
+   * Execution is marked as permanantly failed by its FailureHandlers. Must be manually triggered to
+   * retry.
+   */
+  FAILED,
+
```
kagkarlsson, line 44: Double-check these possibilities are described in the README

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/DeactivatedExecution.java
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+package com.github.kagkarlsson.scheduler;
+
+import com.github.kagkarlsson.scheduler.task.Execution;
+import com.github.kagkarlsson.scheduler.task.State;
+import com.github.kagkarlsson.scheduler.task.TaskInstanceId;
+import java.time.Instant;
+
+public record DeactivatedExecution(
```
kagkarlsson: This is not used, should be removed

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
```diff
@@ -48,27 +50,28 @@ public class Scheduler implements SchedulerClient {
...
   private final Duration deleteUnresolvedAfter;
+  private final Duration deleteDeactivatedAfter;
   private final Duration shutdownMaxWait;
-  protected final List<OnStartup> onStartup;
   private final Waiter detectDeadWaiter;
   private final Duration heartbeatInterval;
-  final SchedulerListeners schedulerListeners;
   private final ExecutorService dueExecutor;
   private final Waiter heartbeatWaiter;
-  final SettableSchedulerState schedulerState = new SettableSchedulerState();
```
kagkarlsson, line 70: Ideally minimal change of ordering of fields.. (use separate refactor PR)

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
```diff
@@ -416,6 +444,35 @@ protected void detectDeadExecutions() {
...
+        EnumSet.of(State.COMPLETE, State.PAUSED, State.FAILED, State.WAITING);
+    int limit = 10000;
+
+    int totalRemoved = 0;
+    int removed;
+    do {
+      removed =
+          schedulerTaskRepository.removeOldDeactivatedExecutions(deleteInStates, olderThan, limit);
+      totalRemoved += removed;
+    } while (removed == limit && !schedulerState.isShuttingDown());
```
kagkarlsson: This might run on all scheduler instances concurrently. Is there a risk for deadlock?

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerClient.java
```diff
@@ -585,9 +609,46 @@ public void cancel(TaskInstanceId taskInstanceId) {
...
+    @Override
+    public void reactivate(TaskInstanceId taskInstanceId, Instant newExecutionTime) {
+      Execution execution =
+          taskRepository
+              .getExecution(taskInstanceId)
+              .orElseThrow(() -> new TaskInstanceNotFoundException(taskInstanceId));
+
+      if (execution.isPicked()) {
+        throw new TaskInstanceCurrentlyExecutingException(taskInstanceId);
+      }
```
kagkarlsson: Probably redundant. Can never be picked in deactivated state, so possibly IllegalStateException if checking this, possibly drop check..

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
```diff
@@ -416,6 +444,35 @@ protected void detectDeadExecutions() {
...
+      LOG.trace("Deletion of old deactivated executions is disabled.");
+      return;
+    }
+
+    LOG.debug("Deleting old deactivated executions.");
+    Instant olderThan = clock.now().minus(deleteDeactivatedAfter);
+
+    // Delete all deactivated states except RECORD (kept indefinitely)
+    EnumSet<State> deleteInStates =
+        EnumSet.of(State.COMPLETE, State.PAUSED, State.FAILED, State.WAITING);
```
kagkarlsson: Think we need to reconsider this. Maybe limit to COMPLETE, FAILED.

* COMPLETE - default end-state after successful run, should not keep forever
* FAILED - might need automatic cleanup, similar to how AWS SQS only keep events 14 days on DLQ
* PAUSED - manual operation to set as PAUSED, should probably not cleanup automatically
* WAITING - use-case specific max wait-time, should probably not cleanup automatically

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
kagkarlsson: A fix might be to enforce an order, e.g. `order by execution_time desc`

kagkarlsson: Or possibly randomize when it runs..

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
kagkarlsson: Also. 
* Consider using a strategy with `select` + `delete where in` to avoid complicated query and support for all databases.
* Consider adding `ORDER BY execution_time desc`
* Consider randomizing time when cleanup runs to avoid deadlocks

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/State.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public enum State {
+  /** Execution is active and scheduled for execution. */
+  ACTIVE,
+
+  /**
+   * Execution completed successfully and is kept for deduplication. Can be deleted by automatic
+   * process after configured time.
+   */
+  COMPLETE,
```
kagkarlsson, line 28: Consider if these should have another representation in the database, e.g. `ACTI`/`COMP`/`FAIL`... or similar

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/task/helper/Tasks.java
kagkarlsson: Additionally, consider adding (or better naming):
```java
.onCompleteKeepAsRecord()
.onCompleteDelayRemoval()
```

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/ScheduledExecutionsFilter.java
```diff
@@ -23,15 +23,33 @@ public class ScheduledExecutionsFilter {
...
+
+  /** Returns a filter for only active (scheduled) executions. This is the recommended default. */
+  public static ScheduledExecutionsFilter active() {
+    return new ScheduledExecutionsFilter(StateTypeFilter.ACTIVE).withIncludeUnresolved(true);
+  }
 
+  /** Returns a filter for only deactivated executions (non-active states). */
+  public static ScheduledExecutionsFilter deactivated() {
+    return new ScheduledExecutionsFilter(StateTypeFilter.DEACTIVATED).withIncludeUnresolved(true);
+  }
```
kagkarlsson, line 40: Should be able to filter by state

PR #801, opened 2026-03-26
refactor: ‎pollUsingFetchAndLockOnExecute strategy is renamed and ‎README is refactored

strelchm:
## Clarify polling strategy naming between documentation and Spring Boot config and additional readme refactoring
The polling strategy is referred to as `fetch-and-lock-on-execute` in the main documentation, but configured as `fetch` in Fluent API and Spring Boot application.properties. 
Specifying `fetch-and-lock-on-execute` in Spring configuration results in startup failure cause the enum expects `FETCH` or `LOCK_AND_FETCH`:
```
***************************
APPLICATION FAILED TO START
***************************

Description:

Failed to bind properties under 'db-scheduler.polling-strategy' to com.github.kagkarlsson.scheduler.PollingStrategyConfig$Type:

    Property: db-scheduler.polling-strategy
    Value: "fetch-and-lock-on-execute"
    Origin: class path resource [application.properties] - 26:31
    Reason: failed to convert java.lang.String to com.github.kagkarlsson.scheduler.PollingStrategyConfig$Type (caused by java.lang.IllegalArgumentException: No enum constant com.github.kagkarlsson.scheduler.PollingStrategyConfig.Type.fetch-and-lock-on-execute)

Action:

Update your application's configuration. The following values are valid:

    FETCH
    LOCK_AND_FETCH
```
This inconsistency can be confusing for users. Changing the property name from `fetch` to `fetch-and-lock-on-execute` would constitute a breaking change for existing users.  Therefore, renaming `fetch-and-lock-on-execute` to `fetch` is the optimal variant  solution — it requires only documentation updates and the deprecation of one scheduler method.

## Fixes

This PR establishes clear consistency between the two:
- Renames polling strategy `fetch-and-lock-on-execute` to `fetch` in `‎README.md` and scheduler builder. For backward compatibility `builder.pollUsingFetchAndLockOnExecute` method is marked as deprecated 
- Adds "SQL Server, MySQL v8+" to the lock-and-fetch description in the Benchmark test. In the pollUsingLockAndFetch config description, these databases are already listed.

Also additional refactoring:
- Removes duplicate polling strategy properties in Spring Boot example
- Replaces corrupted herokuapp Maven Central badge with modern Shields.io badge. Update link to point to central.sonatype.com

Documentation is updated to clarify the mapping, helping users avoid this error.

## Reminders

---

cc @kagkarlsson

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # db-scheduler
 
 ![build status](https://github.com/kagkarlsson/db-scheduler/workflows/build/badge.svg)
-[![Maven Central](https://maven-badges.herokuapp.com/maven-central/com.github.kagkarlsson/db-scheduler/badge.svg)](https://maven-badges.herokuapp.com/maven-central/com.github.kagkarlsson/db-scheduler)
+[![Maven Central](https://img.shields.io/maven-central/v/com.github.kagkarlsson/db-scheduler)](https://central.sonatype.com/artifact/com.github.kagkarlsson/db-scheduler)
 [![License](http://img.shields.io/:license-apache-brightgreen.svg)](http://www.apache.org/licenses/LICENSE-2.0.html)
 
 Task-scheduler for Java that was inspired by the need for a clustered `java.util.concurrent.ScheduledExecutorService` simpler than Quartz.
@@ -276,7 +276,7 @@ as they do not support descending indexes.
 #### Polling strategy
 
 If you are running >1000 executions/s you might want to use the `lock-and-fetch` polling-strategy for lower overhead
-and higher throughput ([read more](#polling-strategy-lock-and-fetch)). If not, the default `fetch-and-lock-on-execute` will be fine.
+and higher throughput ([read more](#polling-strategy-lock-and-fetch)). If not, the default `fetch-and-lock-on-execute` (configured as `fetch` in Fluent API and Spring Boot) will be fine.
 
 :gear: `.pollUsingFetchAndLockOnExecute(double, double)`<br/>
 Use default polling strategy `fetch-and-lock-on-execute`.<br/>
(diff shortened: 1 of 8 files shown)


kagkarlsson:
Thank you! That mismatch is unfortunate, and I remember naming those strategies was not easy. I have not checked, but how hard would it be to harmonize and deprecate the odd-one out?

README.md
```diff
@@ -464,7 +461,7 @@ db-scheduler.priority-enabled=false
 # Ignored if a custom DbSchedulerStarter bean is defined
 db-scheduler.delay-startup-until-context-ready=false
 
-db-scheduler.polling-strategy=fetch
+db-scheduler.polling-strategy=fetch # fetch-and-lock-on-execute (default)
```
kagkarlsson: Comment on separate line

kagkarlsson (approved):

strelchm:
> Thank you! That mismatch is unfortunate, and I remember naming those strategies was not easy. I have not checked, but how hard would it be to harmonize and deprecate the odd-one out?

I think the optimal variant is to change `fetch-and-lock-on-execute` ->  `fetch` cause:

- one point of truth
- clients reading code will naturally think of the default strategy as `fetch` — it's simple and more memorable
- backward compatibility — only one deprecated `builder.pollUsingFetchAndLockOnExecute` method instead of a total breaking change due to the enum value renaming



Other variants:

1. change `fetch` -> `fetch-and-lock-on-execute` in enum PollingStrategyConfig.Type and spring configs. It breaks backward compatibility in code
2. spring enum property Converter - overhead and no single point of truth

kagkarlsson:
Looks good, would you update the PR description?

strelchm:
> Looks good, would you update the PR description?

Done. Updated according to the latest changes.

kagkarlsson:
Sorry, also a merge-conflict before being able to merge

strelchm:
> Sorry, also a merge-conflict before being able to merge

rebased

Issue #804, opened 2026-03-30
Behavior for unresolved tasks differs between polling strategies


strelchm:
In the FETCH polling strategy, when fetching tasks:
```
select * from scheduled_tasks where picked = ? and execution_time <= ? ORDER BY execution_time ASC LIMIT 15
```
if a fetched task has an unresolved task type, we add this type to **unresolvedTasks** and do not process it, including the picked-field update. Other instances can process this task because it is not in the picked state. This is very helpful for rolling updates — when one old instance does not know a new task type, but a newer instance can process it without big delay.

In the LOCK-AND-FETCH strategy, after fetching tasks:
```
WITH locked_executions as (UPDATE scheduled_tasks st1 SET picked = ?, picked_by = ?, last_heartbeat = ?, version = version + 1 WHERE (st1.task_name, st1.task_instance) IN (SELECT st2.task_name, st2.task_instance FROM scheduled_tasks st2 WHERE picked = ? and execution_time <= ? ORDER BY execution_time ASC FOR UPDATE SKIP LOCKED LIMIT 15) RETURNING st1.*) SELECT * FROM locked_executions ORDER BY execution_time ASC
```
the task is in the picked state in the DB. Then, if it has an unresolved task type for the current instance, we add this type to **unresolvedTasks** and finish processing it, as in FETCH mode. But the task stays in the picked state, and other instances cannot process it instantly — they can process it only when it becomes a dead execution, because its heartbeat does not update while in the picked state.

The behavior of the first unresolved task differs between polling strategies, which is unintuitive. Perhaps returning the picked field to false as a compensatory action in LOCK-AND-FETCH mode could allow successful processing of new tasks during rolling updates without waiting for the dead execution interval, which can be long.

kagkarlsson:
Interesting, I was not aware of this. I agree that the behavior should be the same. 
db-scheduler does not technically fully support multiple instances with different task-set using the same table. I am however warming to the idea, see #442

strelchm:
Thanks for response!

I've been considering a quick fix to handle the issue in our project (we frequently observe jobs freezing of several jobs after rolling updates when introducing a new job type).  ~~My idea is to introduce a scheduler listener UnresolvedSchedulerListener - that detects and unpicks unresolved jobs assigned to a current instance. Here's how it would work:~~

 ~~When the scheduler detects unresolved tasks (e.g., tasks picked but not progressing), it emits a UNRESOLVED_TASK event.
[...]

PR #809, opened 2026-04-11
fix: lock-and-fetch unresolved first task picked state is fixed

strelchm:
## Scheduler listener that handles unresolved tasks for LOCK_AND_FETCH mode is added
When a task type becomes unresolved (e.g., not registered in the current scheduler instance due to a rolling update where new versions don't have the task yet), this listener automatically releases (unpicks) those tasks so they can be processed by other scheduler instances that have the task type available (this behavior is already built-in 
for FETCH mode out of the box).

### Implementation Details

**Listener behavior:**
- Added to scheduler in builder only in LOCK_AND_FETCH mode
- Tracks previously unpicked task names to prevent duplicate queries
- Reduces unnecessary database operations

**Unpick operation:**
- Uses batch update to mark tasks as unpicked (picked=false with version increment)
- Records contain picker info (picked_by scheduler instance) and execution_time
- Allows other scheduler instances to pick up unresolved tasks

### Known Issue / Limitation

⚠️ **Current limitation:** If batch contains multiple unresolved **instances** of the same task, only the first instance is unpicked (cause only first unresolved task is recorded in TaskResolver as unresolved). TaskResolver needs to store all unresolved task instances (by taskName + executionTime), not just one per task name to unpick all possible unresolved tasks.

### Tests Added
Several additional tests are added to TaskResolverTest for better coverage.

## Fixes

https://github.com/kagkarlsson/db-scheduler/issues/804

## Reminders

---

cc @kagkarlsson

### DIFF
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/UnresolvedTasksLockAndFetchListener.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/UnresolvedTasksLockAndFetchListener.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright (C) Gustav Karlsson
+ *
+ * <p>Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * <p>http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * <p>Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.kagkarlsson.scheduler;
+
+import com.github.kagkarlsson.scheduler.TaskResolver.UnresolvedTask;
+import com.github.kagkarlsson.scheduler.event.AbstractSchedulerListener;
+import com.github.kagkarlsson.scheduler.jdbc.JdbcTaskRepository;
+import java.util.List;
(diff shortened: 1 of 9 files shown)


db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/UnresolvedTasksLockAndFetchListener.java
```diff
@@ -0,0 +1,78 @@
...
+      final List<String> unresolvedTaskNames =
+          taskResolver.getUnresolved().stream()
+              .map(UnresolvedTask::getTaskName)
+              .filter(taskName -> !refusedTasks.contains(taskName))
+              .toList();
+      unpickUnresolvedTasks(unresolvedTaskNames);
+    }
+  }
+
+  private void unpickUnresolvedTasks(List<String> unresolvedTaskNames) {
```
strelchm, line 63: I can add the execution ids of unresolved tasks to TaskResolver to unpick not only the first unresolved task but also the others if there are more than one unresolved task in first batch.


⚠️ Current limitation: If batch contains multiple unresolved instances of the same task, only the first instance is unpicked (cause only first unresolved task is recorded in TaskResolver as unresolved). TaskResolver needs to store all unresolved task instances (by taskName + executionTime), not just one per task name to unpick all possible unresolved tasks.

strelchm: This second problem is that new listener triggers after every task resolving. The idea is bad and I will implement the other, optimal solution

strelchm:
The idea with listener is bad and I will implement the other, optimal solution in https://github.com/kagkarlsson/db-scheduler/pull/811

PR #811, opened 2026-04-18
fix: single statement lock-and-fetch unresolved task from first batch…

strelchm:
## Added unpicking of unresolved tasks in single‑statement `LOCK_AND_FETCH` mode
When a task type becomes unresolved (e.g., not registered in the current scheduler instance due to a rolling update where new versions don't have the task yet), the compensation batch query unpicks those tasks so they can be processed by other scheduler instances that have the task type available (this behavior is already built-in for `FETCH` and generic `LOCK_AND_FETCH` modes out of the box).

### Implementation Details

**Compensation query in single statement `LOCK_AND_FETCH` mode:**
- Added a transaction. The first query is **existed** `SELECT FOR UPDATE SKIP LOCKED` that picks a task. The second query unpicks unresolved tasks if any exist. If there are no unresolved executions, the second query is skipped and only the transaction commit is performed.
- Allows other scheduler instances to pick up unresolved tasks

**Unpick operation (`unpickPickedBatch` repository method):**
- Uses a batch update to mark tasks as unpicked (`picked = false` with version increment).
If any of the unpicked tasks is not found or has a mismatched version, the method throws an exception.

### Tests Added
Several additional tests are added to `TaskResolverTest` for better coverage.

## Fixes

https://github.com/kagkarlsson/db-scheduler/issues/804

## Reminders

---

cc @kagkarlsson


db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
```diff
@@ -343,6 +346,28 @@ public List<Execution> getDue(Instant now, int limit) {
     return getExecutions(jdbcRunner, selectDueQuery, now, limit, unresolvedFilter);
   }
 
+  private List<Execution> lockAndFetchSingleStatement(Instant now, int limit) {
+    return jdbcRunner.inTransaction(
```
strelchm: There is some tradeoff to add a transaction instead of one query. I think atomicity is better here — the second query will be executed only in rare unresolved task cases, and the transaction impact will be just the second  roundtrip (for batch) .

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
```diff
@@ -343,6 +346,28 @@ public List<Execution> getDue(Instant now, int limit) {
...
+              jdbcCustomization.lockAndFetchSingleStatement(
+                  getTaskRespositoryContext(), now, limit, orderByPriority);
+
+          Map<Boolean, List<Execution>> allCandidatesByIsUnresolved =
+              allCandidates.stream()
+                  .collect(
+                      partitioningBy(
+                          execution -> taskResolver.isUnresolved(execution.getTaskName())));
+
+          List<Execution> unresolvedExecutions = allCandidatesByIsUnresolved.get(true);
```
OptimumCode: nitpick: We could use `Boolean.TRUE` to avoid unnecessary boxing here (and `Boolean.FALSE` below)

OptimumCode (commented):
@strelchm Thanks for making the PR! We are currently experiencing a problem with false-positive dead executions because the lock-and-fetch strategy does not unpick unresolved tasks (when they were grabbed by an old worker that does not yet have the task definition). And this PR is just what we need!

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
strelchm: Good point, thanks! Fixed.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
```diff
@@ -578,6 +602,50 @@ public Optional<Execution> pick(Execution e, Instant timePicked) {
...
+        unpickedTasks.add(task);
+      } else {
+        nonUnpickedTasks.add(task);
+      }
+    }
+    if (!nonUnpickedTasks.isEmpty()) {
+      throw new FailedToUnpickBatchException(
+          "Error while unpicking tasks batch. Failed to unpick: %s, successfully unpicked: %s"
+              .formatted(nonUnpickedTasks, unpickedTasks));
+    }
```
kagkarlsson: Should we perhaps change this to logging an error instead of throwing? (to not block polling)

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/TaskRepository.java
```diff
@@ -86,6 +86,8 @@ boolean reschedule(
 
   Optional<Execution> pick(Execution e, Instant timePicked);
 
+  void unpickPickedBatch(List<Execution> unresolvedExecutions);
```
kagkarlsson: `unpickPickedBatch` -> `unpickUnresolved` ?

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/TaskResolver.java
```diff
@@ -85,6 +85,10 @@ public List<String> getUnresolvedTaskNames(Duration unresolvedFor) {
         .collect(Collectors.toList());
   }
 
+  public boolean isUnresolved(String taskName) {
+    return unresolvedTasks.containsKey(taskName);
+  }
+
```
kagkarlsson, line 91: What if this is the first occurrance, i.e. nothing has tried to resolve it yet, or is that actually guaranteed after a poll? (have not checked myself)

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/JdbcTaskRepositoryTest.java
```diff
@@ -538,6 +552,158 @@ public void lockAndFetchGeneric_happy() {
...
+    List<Execution> picked = taskRepository.lockAndFetchGeneric(now, POLLING_LIMIT);
+    assertThat(picked, hasSize(1));
+    assertTrue(taskResolver.isUnresolved(unknownOneTimeTask.getName()));
+
+    checkPicked(knownTaskInstance, true);
+    checkPicked(unknownTaskInstance, false);
+  }
+
+  @Test
+  public void unpickPickedBatch_should_release_unknown_picked_tasks() {
```
kagkarlsson: Do we need separate tests on `unpickPickedBatch_...` or is it perhaps sufficient with those above?

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/JdbcTaskRepositoryTest.java
```diff
@@ -538,6 +552,158 @@ public void lockAndFetchGeneric_happy() {
...
+    List<Execution> picked = taskRepository.lockAndGetDue(now, POLLING_LIMIT);
+    assertThat(picked, hasSize(1));
+    assertTrue(taskResolver.isUnresolved(unknownOneTimeTask.getName()));
+    assertTrue(taskResolver.isUnresolved(unknownSecondOneTimeTask.getName()));
+
+    checkPicked(knownTaskInstance, true);
+    checkPicked(unknownTaskInstance1, false);
+    checkPicked(unknownOneTimeTaskInstance2, false);
+    checkPicked(unknownAlternativeOneTimeTaskInstance, false);
+  }
```
kagkarlsson: Anything that can be done to reduce boilerplate/lines? (for all tests)
(I know this class has some legacy in that department though....)

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/SchedulerTest.java
```diff
@@ -338,4 +344,56 @@ public void should_reschedule_failure_on_exponential_backoff_with_defined_rate()
           greaterThanOrEqualTo(expectedTimeDifferenceFromFirstCall.minusSeconds(1).getSeconds()));
     }
   }
+
+  @Test
+  public void
+      should_unpick_unresolved_tasks_during_rolling_update_in_single_statement_lock_and_fetch() {
+    final TestTasks.CountingHandler<Void> knownTaskHandler = new TestTasks.CountingHandler<>();
```
kagkarlsson: Nor sure we need tests at this level, what are your thoughts? (trying to avoid adding unnecessary tests)

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/TaskResolverTest.java
```diff
@@ -30,4 +37,72 @@ void shouldProduceUnresolvedTaskEventWhenTheTaskResolverIsUnableToResolveTheTask
...
+    taskResolver.resolve(
+        new Execution(
+            now.plus(Duration.ofSeconds(10)), new TaskInstance<Void>("unresolved2", "id2")));
+
+    final List<UnresolvedTask> unresolved = taskResolver.getUnresolved();
+    assertThat(unresolved, hasSize(2));
+    assertThat(
+        unresolved.stream().map(TaskResolver.UnresolvedTask::getTaskName).toList(),
+        contains("unresolved1", "unresolved2"));
+  }
```
kagkarlsson, line 43: Again, anything that can be done to reduce duplication and verbosity?

kagkarlsson (commented):
Looks like good fix for the issue, great job 👍 
I just have a few comments

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/TaskRepository.java
strelchm: I thought to do so, but I'd like to make this method universal — not just for `unresolved` tasks, but for picking/locking in other situations as well. The term `unresolved` is unclear in the DAO context.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/TaskResolver.java
strelchm: - This method is called after `JdbcCustomization.lockAndFetchSingleStatement` in `JdbcTaskRepository#lockAndFetchSingleStatement`.
- If the task is unresolved - it will be added to the `TaskResolver#unresolvedTasks` method in `JdbcCustomization.lockAndFetchSingleStatement` (during record mapping in `ExecutionResultSetConsumer#map`). 

In a single statement lock and fetch, it's guaranteed that an unresolved task will be in the `unresolvedTasks` filter when "unpicking" check occurs.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
strelchm: - I think we should not change this, because an unexpected return count should not occur — the rows are locked by `SKIP LOCKED`, and nothing can prevent it from updating to unpicked and returning a result different from 1. 
- If there is an error in a batch statement, then a `BatchUpdateException` should be thrown.

Exception throwing differs from generic lock and fetch, which only [logs that some tasks failed to be picked](https://github.com/kagkarlsson/db-scheduler/blob/master/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java#L385). But in that mode if a task is not picked, it's not bad — another instance could run it in generic lock and fetch mode. Logging is optimal.

However, if a task is not unpicked in single statement mode, it would be frozen for more than five minutes before the dead execution handler runs. This can be bad for tasks that are critical to freeze according to business logic. With exception transaction rollback and retry means tasks are not picked and could run in this or another instance.

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/JdbcTaskRepositoryTest.java
strelchm: I have deleted one duplicated test since your comment, but the others are needed

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/JdbcTaskRepositoryTest.java
strelchm: I've removed several tests and compacted some of them into a parameterized test

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/SchedulerTest.java
strelchm: I think that one integration test for the feature of unresolved tasks is needed — to cover the bug situation. Changed the test to parameterized for both modes.

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/TaskResolverTest.java
strelchm: refactored

strelchm:
> I just have a few comments

I have refactored tests/code and added a description to README.md in a separate commit.

@kagkarlsson, @OptimumCode, please check when you have time.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
OptimumCode: I think this query will also be executed every time when a previously existing task definition was removed. Meaning during `removeUnresolvedAfter` period, this query will be executed on every fetch attempt (because there will be a record in db for an old task definition). Please correct me, if I am wrong here.

But still, the overall performance impact should not be significant, I guess

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/JdbcTaskRepositoryTest.java
```diff
@@ -538,6 +555,85 @@ public void lockAndFetchGeneric_happy() {
...
+    TaskInstance<?> unknownAlternativeTaskInstance =
+        unknownSecondOneTimeTask.instance("unknownAlternativeTaskInstance");
+
+    createIfNotExists(
+        SchedulableInstance.of(knownTaskInstance, now),
+        SchedulableInstance.of(unknownInstance1, now),
+        SchedulableInstance.of(unknownInstance2, now),
+        SchedulableInstance.of(unknownAlternativeTaskInstance, now));
+
+    List<Execution> picked;
```
OptimumCode: suggestion: we could use a method source to get rid of the if statement inside test.

Personally, I always try to avoid any if statements in the test (whenever you have an if, it becomes harder to understand what test evaluates). We could use method source with lambda to simplify the test body. It would look like this:

```java

@ParameterizedTest
@MethodSource("fetchMethods")
void should_release_unknown_tasks(Function<Instant, List<Execution>> taskSupplier, boolean pickKnownTasks) {
  //...
  List<Execution> picked = taskSupplier.apply(now);
  //...
}

static void fetchMethods(): List<Arguments> {
  Function<Instant, List<Execution>> getDue = now -> taskRepository.getDue(now, POLLING_LIMIT);
  // other methods
  return List.of(
    Arguments.argumentSet("getDue", getDue, false),
    // other sets
  )
}

```

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/JdbcTaskRepositoryTest.java
```diff
@@ -538,6 +555,85 @@ public void lockAndFetchGeneric_happy() {
...
+        ex.getMessage());
+  }
+
+  @Test
+  public void unpickPickedBatch_should_handle_empty_collection_gracefully() {
+    assertDoesNotThrow(() -> taskRepository.unpickPickedBatch(Collections.emptyList()));
+  }
+
+  @Test
+  void unpickPickedBatch_should_not_crash_when_Picked_batch_task_not_in_database() {
```
OptimumCode: nitpick: Inconsistent method modifiers

It looks like all other methods in this class have a `public` modifier. Probably, it is better to keep consistent modifiers across all test methods

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/TaskResolverTest.java
```diff
@@ -1,33 +1,83 @@
...
-import com.github.kagkarlsson.scheduler.task.Execution;
-import com.github.kagkarlsson.scheduler.task.TaskInstance;
 import java.time.Instant;
 import java.util.List;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
-import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 
 @ExtendWith(MockitoExtension.class)
```
OptimumCode: question: Do we still need this extension, after we have removed `@Mock` annotation?

strelchm: removed

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/JdbcTaskRepositoryTest.java
strelchm: made this method public

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/JdbcTaskRepositoryTest.java
strelchm: I thought to pass Scheduller as argument, your suggestion is good. I will fix later

strelchm: https://github.com/kagkarlsson/db-scheduler/compare/4ee850078e2146632b8052f63d635a0b65ea7cd2..ca73ef5c40726a1505fc17d576451f1955f57067
Fixed two methods. Besides if-s I forgot to change method name, fixed. Thank you!

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
kagkarlsson: I am not sure it is a big deal that the execution is frozen for 5mins, since there is no defined implementation for the task, so nothing should be waiting to run it. _Unless_ we are running multiple Schedulers with two different task-sets against the same table, which currently is not explicitly supported, but is a requested thing..

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
kagkarlsson: I think `inTransaction` will issue a `setAutocommit(false)` which will roundtrip to the db, so there is that if very concerned about polling-time. 
Not sure atomicity is worth it since it should be a really rare occurrence (unless again we are running multiple schedulers with different task-sets, for which there are better "fixes")

strelchm: Yes, I mean the round-trip caused by the commit in transaction close, as @kagkarlsson mentioned (The second query will be executed only rarely, but the commit round-trip will occur on every poll). Now I agree that the impact of nonatomicity is relatively small compared to the performance cost of the additional round-trip. Even without a transaction, the task row cannot be altered, as the task is already picked and remains unchanged by the API or other background processes. Therefore, I’ll remove the transaction.

> unless again we are running multiple schedulers with different task-sets, for which there are better "fixes"

This could also occur during a rolling update or canary deployment, where two pod versions may coexist temporarily (our example: more then 10 pods and new task release, rolling update in k8s).

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
strelchm: >  Unless we are running multiple Schedulers with two different task-sets against the same table, which currently is not explicitly supported, but is a requested thing..

This could also occur during a rolling update or canary deployment, where two pod versions may coexist temporarily (our example: more then 10 pods and new task release, rolling update in k8s).

strelchm: - In the case of this method:
  - Version updates cannot occur, because the task is picked only after the SKIP LOCKED query, ensuring it is not reprocessed (even if this two queries are not in transaction).
  - If there is another error in the batch statement, a BatchUpdateException should be thrown.
- In other cases the exception can be useful.

However, if you believe it would be safer to log the error instead of throwing an exception, we can adjust the behavior accordingly.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
strelchm: Transaction is removed in [separate commit](https://github.com/kagkarlsson/db-scheduler/pull/811/changes/bfa153dcbf980d5bbbf13a161aac9303ff02f10a)

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
kagkarlsson, line 378: This code does not need to unpick since they are already filtered out via `getExecutions` ?

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
```diff
@@ -343,6 +346,24 @@ public List<Execution> getDue(Instant now, int limit) {
...
+            .collect(
+                partitioningBy(execution -> taskResolver.isUnresolved(execution.getTaskName())));
+
+    List<Execution> unresolvedExecutions = allCandidatesByIsUnresolved.get(Boolean.TRUE);
+    if (unresolvedExecutions != null && !unresolvedExecutions.isEmpty()) {
+      unpickPickedBatch(unresolvedExecutions);
+    }
+
+    return allCandidatesByIsUnresolved.get(Boolean.FALSE);
+  }
```
kagkarlsson, line 364: Better 👍

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
```diff
@@ -578,6 +598,50 @@ public Optional<Execution> pick(Execution e, Instant timePicked) {
...
+    final List<Execution> unpickedTasks = new ArrayList<>();
+    final List<Execution> nonUnpickedTasks = new ArrayList<>();
+    for (int i = 0; i < updated.length; i++) {
+      Execution task = executions.get(i);
+      if (updated[i] == 1) {
+        unpickedTasks.add(task);
+      } else {
+        nonUnpickedTasks.add(task);
+      }
+    }
```
kagkarlsson, line 635: See 
`public List<Execution> lockAndFetchGeneric(Instant now, int limit) {`

Is it possible to create a reusable method for batch-updates for fetting any executions that were not updated?

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
kagkarlsson: Lets log error rather than throw for now.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/TaskRepository.java
kagkarlsson: Ok, that is fine 👍. Maybe rename parameter

README.md
```diff
@@ -557,6 +554,16 @@ JVM running the scheduler suddenly exits.
...
+### Unknown tasks
+
+If a task instance is found in the database but the corresponding task definition is not registered in the service (e.g., during due execution or dead execution housekeeping), it is treated as an unknown task (see [`TaskResolver`](https://github.com/kagkarlsson/db-scheduler/blob/master/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/TaskResolver.java)).
+
+Behavior of unknown tasks:
+
+* They are excluded from polling — the current instance will not attempt to pick or execute them.
+* They remain in the database so other instances (e.g., newer versions in a rolling update or canary deployment) can pick and process them.
+* They are **automatically removed** after a configured retention period `deleteUnresolvedAfter`, if they remain unresolved.
+
```
kagkarlsson: I think I have called these "Unresolved tasks", no? But nice with a doc-entry for it 👍

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/JdbcTaskRepositoryTest.java
```diff
@@ -538,6 +556,90 @@ public void lockAndFetchGeneric_happy() {
...
+    List<Execution> picked = taskSupplier.apply(now, taskRepository);
+    assertThat(picked, hasSize(1));
+    assertTrue(taskResolver.isUnresolved(unknownOneTimeTask.getName()));
+    assertTrue(taskResolver.isUnresolved(unknownSecondOneTimeTask.getName()));
+
+    checkPicked(lockAndFetch, knownTaskInstance);
+    checkPicked(false, unknownInstance1, unknownInstance2, unknownAlternativeTaskInstance);
+  }
+
+  static List<Arguments> fetchMethodsAndIsLockAndFetch() {
```
kagkarlsson: Naming?

kagkarlsson (commented):
Updated review

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/TaskRepository.java
strelchm: Renamed to `pickedExecutions`

README.md
strelchm: By the way, [here](https://github.com/kagkarlsson/db-scheduler/blob/master/README.md?plain=1#L337) in README.md they called "unknown", fixed it too

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
strelchm: This code does not need to unpick because:
- `getExecutions` flter the unknown executions
- `SKIP LOCKED` query does not pick executions
- After this code only known tasks are picked

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
strelchm: This method is ready to be reusable:
- Name and signature are reusable
- Universal mechanism of locking — only picked tasks without version changes will be unpicked
- After refactoring, fail-fast: only possible executions will be unpicked

I wrote this method based on the lockAndFetchGeneric example

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/JdbcTaskRepositoryTest.java
strelchm: Made simpler: `pickFunctions` with property name refactoring

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
strelchm: Done and updated the tests

strelchm:
> Updated review

Fixed according to comments, resolved old comments where I had written answers and made corrections.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
```diff
@@ -578,6 +597,52 @@ public Optional<Execution> pick(Execution e, Instant timePicked) {
...
+    for (int i = 0; i < updated.length; i++) {
+      Execution task = pickedExecutions.get(i);
+      if (updated[i] == 1) {
+        unpickedTasks.add(task);
+      } else {
+        nonUnpickedTasks.add(task);
+      }
+    }
+    if (nonUnpickedTasks.isEmpty()) {
+      LOG.info("Successfully unpicked tasks: {}", unpickedTasks);
```
kagkarlsson: Didn't catch this previously, but prob lower to `debug` since it may happen during normal operations?

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/JdbcTaskRepositoryTest.java
```diff
@@ -538,6 +560,94 @@ public void lockAndFetchGeneric_happy() {
...
+    BiFunction<Instant, JdbcTaskRepository, List<Execution>> lockAndGetDue =
+        (now, taskRepo) -> taskRepo.lockAndGetDue(now, POLLING_LIMIT);
+    return List.of(
+        Arguments.of(getDue, false),
+        Arguments.of(lockAndFetchGeneric, true),
+        Arguments.of(lockAndGetDue, true));
+  }
+
+  @Test
+  public void unpickPickedBatch_should_throw_exception_if_one_of_tasks_is_not_picked() {
```
kagkarlsson: nit: name mismatches now

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/JdbcTaskRepositoryTest.java
```diff
@@ -538,6 +560,94 @@ public void lockAndFetchGeneric_happy() {
...
+  }
+
+  @Test
+  public void unpickPickedBatch_should_handle_empty_collection_gracefully() {
+    assertDoesNotThrow(() -> taskRepository.unpickPickedBatch(Collections.emptyList()));
+  }
+
+  @Test
+  public void unpickPickedBatch_should_not_crash_when_Picked_batch_task_not_in_database() {
+    ListAppender<ILoggingEvent> appender = startAndGetLogListAppender();
```
kagkarlsson: remove appender on finally? (or similar)

kagkarlsson (approved):
Great job on finding and fixing this bug, thanks! 
Just a few minor comments

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
strelchm: fixed

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/JdbcTaskRepositoryTest.java
strelchm: fixed

db-scheduler/src/test/java/com/github/kagkarlsson/scheduler/JdbcTaskRepositoryTest.java
strelchm: fixed

PR #815, opened 2026-04-25
refactor:  reduce ScheduledExecution constructor visibility and add factory method

### LINKED ISSUE #810: refactor: change SchedulerExecution to not expose Execution (via constructor)
Convert into record if possible, replicating the fields needed from Execution

UdayHE:
## Refactor

 ## Converts `ScheduledExecution` to a record, removing `Execution` from the public constructor
                                                                    
  ## Fixes                                                                                                                                                                                                                         
                                                                                                                                                                                                                                   
  Fixes #810      

  - Hides the public constructor (now package-private) to prevent external construction
  - Adds ScheduledExecution.from(Class<T>, Execution) as the public factory method
  - Updates all internal call sites to use the factory method
  - No API disruption, no deprecation churn, equals/hashCode behavior unchanged                                                                                                                                                                                                                 
                                                                                                                                                                                                                                   
  ## Reminders                                                                                                                                                                                                                     

---

cc @kagkarlsson

### DIFF
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/ScheduledExecution.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/ScheduledExecution.java
@@ -20,65 +20,78 @@ import com.github.kagkarlsson.scheduler.task.TaskInstanceId;
 import java.time.Instant;
 import java.util.Objects;
 
-public class ScheduledExecution<DATA_TYPE> {
-  private final Class<DATA_TYPE> dataClass;
-  private final Execution execution;
+@SuppressWarnings("rawtypes")
+public record ScheduledExecution<DATA_TYPE>(
+    Class<DATA_TYPE> dataClass,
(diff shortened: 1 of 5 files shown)


UdayHE:
@kagkarlsson please approve the workflows

kagkarlsson:
I have become uncertain that the record is the best choice. (equals/hashcode override, public constructor and needing to deprecate old methods that now have record-variants)

Would it be better to just reduce visibility of original constructor?

UdayHE:
Agree.
Reduce constructor visibility: This is minimal change, no API disruption, existing equals/hashCode behaviour stays predictable, and no deprecation churn for users.                                                     
I think this is simpler path & probably the right call here. 
Happy to update the PR to just restrict the constructor visibility instead.

kagkarlsson:
Ok, feel free to update the PR 👍

UdayHE:
@kagkarlsson please review

PR #901, opened 2026-07-14
Validate that scheduled_tasks table exists on scheduler startup

### LINKED ISSUE #507: Feature: Validate required table on startup of scheduler
Validate that the required table `scheduled_tasks` exists on startup (in "main-thread"), and possibly also validate expected columns. Fail-fast mode, throw Exception if not, causing the application to not start.

Sharan-0-dot:
Fixes #507

Adds a fail-fast check that validates the configured table (`scheduled_tasks` by
default) exists and has the expected columns when `Scheduler.start()` is called.
If the table is missing or missing a column, the scheduler now throws an
`IllegalStateException` immediately instead of failing later/silently during
polling.

**Implementation:**
- Added `validateTableExists()` to the `TaskRepository` interface, mirroring the
  existing `verifySupportsLockAndFetch()` pattern.
- Implemented in `JdbcTaskRepository` using a lightweight `select ... where 1 = 0`
  query against all expected columns (including `priority` when enabled). This
  avoids per-database `DatabaseMetaData`/table-name-casing special-casing, since
  the query fails identically across all supported databases if the table or a
  column is missing.
- Called at the very start of `Scheduler.start()`, before any threads are started.
- Added `scheduler_should_fail_to_start_if_table_missing` to `SchedulerTest`,
  which points a `SchedulerBuilder` at a nonexistent table name and asserts
  `start()` throws.

**Open question:** this currently runs unconditionally (matching how
`verifySupportsLockAndFetch()` runs unconditionally for the lock-and-fetch
strategy), rather than being opt-in. Happy to add a `.skipTableValidation()`
escape hatch on `SchedulerBuilder` if you'd prefer it not be on by default —
let me know and I'll adjust.

Fixes #507

#### Reminders
- [x] If AI tools were used, disclosed per the [AI Contribution Policy](../AI_POLICY.md)

Used Claude to help me understand the codebase and think through the
implementation approach, as this is my first open-source contribution. I've
reviewed, tested, and understand all the changes and can explain/maintain them.

---
cc @kagkarlsson

### DIFF
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
@@ -93,6 +93,7 @@ public class Scheduler implements SchedulerClient {
       ScheduledExecutorService housekeeperExecutor) {
     this.clock = clock;
     this.schedulerTaskRepository = schedulerTaskRepository;
+    schedulerTaskRepository.validateTableExists();
     this.taskResolver = taskResolver;
     this.threadpoolSize = threadpoolSize;
     this.executor = new Executor(executorService, clock);
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
@@ -20,6 +20,7 @@ import static java.util.stream.Collectors.partitioningBy;
 import static java.util.stream.Collectors.toList;
 
 import com.github.kagkarlsson.jdbc.JdbcRunner;
+import com.github.kagkarlsson.jdbc.PreparedStatementSetter;
 import com.github.kagkarlsson.jdbc.ResultSetMapper;
 import com.github.kagkarlsson.jdbc.SQLRuntimeException;
 import com.github.kagkarlsson.scheduler.Clock;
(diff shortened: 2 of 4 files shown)


db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
```diff
@@ -1,1137 +1,1159 @@
-/*
- * Copyright (C) Gustav Karlsson
```
kagkarlsson, line 2: Why so much whitespace diff?

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
```diff
@@ -1,484 +1,485 @@
...
+    LOG.info("Using polling-strategy: " + pollingStrategyConfig.describe());
+  }
+
+  public void registerSchedulerListener(SchedulerListener listener) {
+    schedulerListeners.add(listener);
+  }
+
+  public void start() {
+    LOG.info("Starting scheduler.");
+    schedulerTaskRepository.validateTableExists();
```
kagkarlsson: Wondering if this could be done in Scheduler constructor...  Ideally you also want a spring boot app to crash, not sure we achieve this if we put it here

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/jdbc/JdbcTaskRepository.java
Sharan-0-dot: Good catch — that was a CRLF/LF line-ending issue from my editor on Windows. I reset the branch against current upstream/master and reapplied just the real changes, so the diff should be clean now.

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
Sharan-0-dot: Moved validateTableExists() into the Scheduler constructor instead of start(), so it now fails during bean construction — should properly crash a Spring Boot app on startup rather than just failing later. Updated the test to assert on .build() instead of .start() accordingly.

Sharan-0-dot:
Pushed an update addressing both comments — rebased against latest upstream/master to fix the line-ending issue, and moved table validation into the Scheduler constructor. mvn clean test passes locally. Ready for another look whenever you have time.

PR #910, opened 2026-08-21
Renaming .instanceWithId to .instance in README

holgerl:
There seem to be a typo in the README.md. 

.instance() takes an id. And there is no method .instanceWithId()

No AI has been used in this PR.

cc @kagkarlsson


holgerl:
Thank you for the merge. I noticed the workflow checks for PR title prefix, but this is not mentioned in the CONTRIBUTING.md guidelines. 

Even though they mention that GA workflows must run, an initial failure could be avoided by mentioning the prefix rule.

PR #913, opened 2026-08-29
fix: validate that pollingInterval is greater than zero

### LINKED ISSUE #339: Validation error for setting pollingInterval=0
Validation error for setting pollingInterval=0

ohchanKyu:
`pollingInterval()` accepted zero and negative durations, 
which makes the polling loop spin without pausing. It is now validated in the builder, like
`missedHeartbeatsLimit()` already is.

Fixes #339

#### Reminders

- [ ] If AI tools were used, disclosed per the [AI Contribution Policy](../AI_POLICY.md)

---

cc @kagkarlsson

### DIFF
--- a/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerBuilder.java
+++ b/db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerBuilder.java
@@ -99,6 +99,12 @@ public class SchedulerBuilder {
   }
 
   public SchedulerBuilder pollingInterval(Duration pollingInterval) {
+    if (pollingInterval == null) {
+      throw new IllegalArgumentException("Polling interval must not be null");
+    }
+    if (pollingInterval.isZero() || pollingInterval.isNegative()) {
+      throw new IllegalArgumentException("Polling interval must be greater than 0");
+    }
     this.poolingInterval = pollingInterval;
     return this;
   }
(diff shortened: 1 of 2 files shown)


db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerBuilder.java
```diff
@@ -99,6 +99,12 @@ public <T extends Task<?> & OnStartup> SchedulerBuilder startTasks(List<T> start
   }
 
   public SchedulerBuilder pollingInterval(Duration pollingInterval) {
+    if (pollingInterval == null) {
+      throw new IllegalArgumentException("Polling interval must not be null");
+    }
+    if (pollingInterval.isZero() || pollingInterval.isNegative()) {
+      throw new IllegalArgumentException("Polling interval must be greater than 0");
+    }
```
kagkarlsson:
```
if (pollingInterval == null || pollingInterval.toMillis() <= 0) {
    ...
}
```
Less lines and reads better?

kagkarlsson (commented):
Thank you for contributing! Small comment

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerBuilder.java
ohchanKyu: Thanks, good point - I’ve applied the change. 
The null case now uses the same message as well.

PR #852, opened 2026-05-16
feat: micrometer executor metrics are added

strelchm:
## feat: micrometer executor metrics are added
New method `registerExecutorsMetrics` is added in `SchedulerBuilder`. This flag adds possibility of registration executor services metrics (does not record timings on operations, only counters). The metrics are similar to Spring schedulers metrics.

By default, the flag is **false**, only default Spring scheduller is in metrics in `spring-boot-example`:
```
# HELP executor_active_threads The approximate number of threads that are actively executing tasks
# TYPE executor_active_threads gauge
executor_active_threads{name="applicationTaskExecutor"} 0.0
# HELP executor_completed_tasks_total The approximate total number of tasks that have completed execution
# TYPE executor_completed_tasks_total counter
executor_completed_tasks_total{name="applicationTaskExecutor"} 0.0
# HELP executor_pool_core_threads The core number of threads for the pool
# TYPE executor_pool_core_threads gauge
executor_pool_core_threads{name="applicationTaskExecutor"} 8.0
# HELP executor_pool_max_threads The maximum allowed number of threads in the pool
# TYPE executor_pool_max_threads gauge
executor_pool_max_threads{name="applicationTaskExecutor"} 2.147483647E9
# HELP executor_pool_size_threads The current number of threads in the pool
# TYPE executor_pool_size_threads gauge
executor_pool_size_threads{name="applicationTaskExecutor"} 0.0
# HELP executor_queue_remaining_tasks The number of additional elements that this queue can ideally accept without blocking
# TYPE executor_queue_remaining_tasks gauge
executor_queue_remaining_tasks{name="applicationTaskExecutor"} 2.147483647E9
# HELP executor_queued_tasks The approximate number of tasks that are queued for execution
# TYPE executor_queued_tasks gauge
executor_queued_tasks{name="applicationTaskExecutor"} 0.0
```
After `registerExecutorsMetrics` enabling the metrics looks like:
```
# TYPE executor_active_threads gauge
executor_active_threads{name="applicationTaskExecutor"} 0.0
executor_active_threads{name="dbSchedulerExecutor"} 0.0
executor_active_threads{name="dbSchedulerHousekeeperExecutor"} 0.0
# HELP executor_completed_tasks_total The approximate total number of tasks that have completed execution
# TYPE executor_completed_tasks_total counter
executor_completed_tasks_total{name="applicationTaskExecutor"} 0.0
executor_completed_tasks_total{name="dbSchedulerExecutor"} 2.0
[...]

PR #915, opened 2026-08-29
feat: long running executions logging is added

strelchm:
## Long running executions logging housekeeping job is added
- A new periodic job that runs in the housekeeping executor.
- The `log-long-running-executions` property is added (default value: true).
- The `long-running-executions-logging-threshold` property is added (default value: 30 minutes).
- The initial delay of the job is the value of the `long-running-executions-logging-threshold` property.
- The job delay is the minimum of `heartbeatInterval * 2` and `longRunningExecutionsThresholdMillis`.
- A warning message is logged with the execution duration and thread stacktrace.

Configuration example:
```
db-scheduler.log-long-running-executions=true
db-scheduler.long-running-executions-logging-threshold=30m
```
Example with the configuration above:
```
2026-08-29T19:54:44.764+07:00  INFO 57104 --- [spring-boot-example] [           main] o.s.boot.tomcat.TomcatWebServer          : Tomcat started on port 8090 (http) with context path '/'
2026-08-29T19:54:44.773+07:00  INFO 57104 --- [spring-boot-example] [           main] c.github.kagkarlsson.examples.boot.App   : Started App in 6.004 seconds (process running for 6.922)
2026-08-29T19:54:44.852+07:00  INFO 57104 --- [spring-boot-example] [pool-2-thread-2] c.g.k.e.b.c.BasicExamplesConfiguration   : One-time long running task starting...
2026-08-29T20:24:46.476+07:00  WARN 57104 --- [spring-boot-example] [pool-4-thread-2] c.g.kagkarlsson.scheduler.Scheduler      : Execution with TaskInstance: task=sample-one-time-task, id=command-line-runner, priority=0 is long-running (execution time:  PT1801.6229293S). If this message is a false positive and the task is genuinely long-running, try increasing longRunningExecutionsLoggingThreshold or disabling logLongRunningExecutions. Current thread stacktrace: 
		java.base/java.lang.Thread.sleepNanos0(Native Method)
		java.base/java.lang.Thread.sleepNanos(Thread.java:482)
		java.base/java.lang.Thread.sleep(Thread.java:513)
		com.github.kagkarlsson.examples.boot.config.BasicExamplesConfiguration.lambda$sampleOneTimeTask$0(BasicExamplesConfiguration.java:75)
		com.github.kagkarlsson.scheduler.task.helper.Tasks$OneTimeTaskBuilder$1.executeOnce(Tasks.java:303)
		com.github.kagkarlsson.scheduler.task.helper.OneTimeTask.execute(OneTimeTask.java:68)
		com.github.kagkarlsson.scheduler.event.ExecutionChain.proceed(ExecutionChain.java:37)
		com.github.kagkarlsson.scheduler.ExecutePicked.executePickedExecution(ExecutePicked.java:113)
		com.github.kagkarlsson.scheduler.ExecutePicked.run(ExecutePicked.java:89)
		com.github.kagkarlsson.scheduler.FetchCandidates.lambda$run$2(FetchCandidates.java:117)
		java.base/java.util.Optional.ifPresent(Optional.java:178)
		com.github.kagkarlsson.scheduler.FetchCandidates.lambda$run$1(FetchCandidates.java:103)
		com.github.kagkarlsson.scheduler.Executor.lambda$addToQueue$0(Executor.java:53)
		java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1095)
		java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:619)
		java.base/java.lang.Thread.run(Thread.java:1447)
```

## Fixes
https://github.com/kagkarlsson/db-scheduler/issues/116

#### Reminders

- [ ] If AI tools were used, disclosed per the [AI Contribution Policy](../AI_POLICY.md)

---

cc @kagkarlsson


db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/SchedulerBuilder.java
```diff
@@ -52,6 +52,8 @@ public class SchedulerBuilder {
           PollingStrategyConfig.Type.FETCH, 0.5, UPPER_LIMIT_FRACTION_OF_THREADS_FOR_FETCH);
   public static final LogLevel DEFAULT_FAILURE_LOG_LEVEL = LogLevel.WARN;
   public static final boolean LOG_STACK_TRACE_ON_FAILURE = true;
+  public static final boolean LOG_STUCK_EXECUTIONS = true;
+  public static final Duration DEFAULT_STUCK_EXECUTIONS_LOGGING_THRESHOLD = Duration.ofMinutes(30);
```
kagkarlsson: Considering if we should increase this default threshold a bit, to avoid starting to warn for legit long-running tasks. Is 8 hours too long? 

Also suggest in the WARN that if this is a task with a legitimately long execution-time, then the users should increase the threshold or disable stuck-thread logging.

strelchm: In a separate commit:
- Increased the threshold to 8 hours.

In our case, executions are considered stuck after 30 minutes because we have short-running tasks. However, I agree that 8 hours is a reasonable limit where the probability of a false positive is low.
- Updated the log message and example in this MR description.

kagkarlsson:
Nice job picking up this feature-suggestion! I assume you also see value for your own usage of db-scheduler.
How do you feel about changing the term from "stuck executions" to "long-running executions"? (or other suggestions?) My thought was that they might not be stuck, just long-running..

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
```diff
@@ -436,6 +453,33 @@ void updateHeartbeats() {
...
+    if (stuckExecutions.isEmpty()) {
+      LOG.trace("No stuck executions to log.");
+      return;
+    }
+
+    LOG.debug("Logging {} stuck executions being processed.", stuckExecutions.size());
+    stuckExecutions.forEach(
+        execution -> {
+          LOG.warn(
+              "Execution with {} has been stuck for {}. If this message is a false positive and the task is genuinely long-running, try increasing stuckExecutionsLoggingThreshold or disabling logStuckExecutions: {}",
```
kagkarlsson: Maybe `Current thread stacktrace: {}` or something at the end, explaining the stacktrace output?

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
```diff
@@ -436,6 +453,33 @@ void updateHeartbeats() {
...
+            .toList();
+    if (stuckExecutions.isEmpty()) {
+      LOG.trace("No stuck executions to log.");
+      return;
+    }
+
+    LOG.debug("Logging {} stuck executions being processed.", stuckExecutions.size());
+    stuckExecutions.forEach(
+        execution -> {
+          LOG.warn(
```
kagkarlsson, line 473: Should we guard against warning multiple times?

strelchm: - I edited the period of long running executions logging task
- Now it runs every 20 minutes (minimum of 4 heartbeats and threshold 8 hours). I think there no sense to make some job duration limit (additional parameter), because there are no so many logs to make disk or bad user experience problems
- I think that separate log per long running execution is normal because every log contains stacktrace

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
strelchm: Good idea, fixed

strelchm:
> I assume you also see value for your own usage of db-scheduler.

Yes, this feature is good for us because we have many jobs and need additional monitoring (like [this PR](https://github.com/kagkarlsson/db-scheduler/pull/852)). These logs will help us in the same way as HikariCP logs with stacktraces that were useful in some situations

> How do you feel about changing the term from "stuck executions" to "long-running executions"? (or other suggestions?) My thought was that they might not be stuck, just long-running..

I made "long running", then renamed to "stuck". Now I agree that long running is better term for library client because not every long runing job is stuck. Renamed

db-scheduler/src/main/java/com/github/kagkarlsson/scheduler/Scheduler.java
kagkarlsson: Not sure we understand each other.

For a long-running execution, will there not be duplicate ~identical messages every 20min until the scheduler is stopped or the execution is done?

strelchm: Every 20 minutes the message about long-running job will be repeated, so there will be duplicated messages. The user can change threshold according to log

Otherwise, we need to enter the configuration lists of exceptions (long-running jobs).

kagkarlsson: I was thinking that we log at max 1 message per <duration> (or ever) to avoid spamming the logs. 
Of course, the benefit of logging multiple times (every 20min) is that you can see from the logs if it is done or still running.. 

Might be a future improvement as well, if users complain :)

kagkarlsson: > Otherwise, we need to enter the configuration lists of exceptions (long-running jobs).

In theory you could also record that you have logged it on a field on CurrentlyExecuting (I think)

strelchm: > I was thinking that we log at max 1 message per (or ever) to avoid spamming the logs.

In our case of short-lived jobs it's normal to log every 20 minutes (and I think about stuck jobs but not long-running). I understand the concern about spamming.

I can do logging once per job:
- There will be a collection of logged job IDs for logging once
- This collection will be cleared after a certain period (e.g., one day) from this new job

kagkarlsson: Ok, you can try that. Consider if that is better than recording a last-long-running-logged timestamp

PR #916, opened 2026-08-29
feat: replace DbSchedulerCustomizer with DbSchedulerOverrides

kagkarlsson:
Replaces the `DbSchedulerCustomizer` interface with `DbSchedulerOverrides`, an immutable value object with a nested builder. Configuring the starter programmatically no longer means implementing an interface, and an overrides object can be built up conditionally or copied and adjusted via `toBuilder()`.

- New `DbSchedulerOverrides` in the Spring common module: nested builder, `toBuilder()`, `Optional` accessors, null-rejecting setters
- Adds a `customizeBuilder(Consumer<SchedulerBuilder>)` escape hatch, applied last, so it can override property-derived settings and reach `SchedulerBuilder` options the starter does not surface
- `DbSchedulerCustomizer` is deprecated rather than deleted. An existing bean is still bridged with a warning, and overrides win if both are present
- Both Boot 3 and Boot 4 auto-configurations resolve the two types via `ObjectProvider`
- README and the Spring Boot example migrated to the new shape

**Breaking**: the starter no longer registers a no-op `DbSchedulerCustomizer` bean, so injecting that type directly will fail to resolve.

---
This PR was prepared with Claude Code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -452,7 +452,7 @@ For Spring Boot applications, there is a starter `db-scheduler-spring-boot-start
 
 ### Configuration options
 
-Configuration is mainly done via `application.properties`. Settings that cannot be expressed as properties — scheduler-name, serializer, executor-services, `JdbcCustomization` and `DataSource` — are configured by adding a bean of type `DbSchedulerOverrides` to your Spring context.
+Configuration is mainly done via `application.properties`. Settings that cannot be expressed as properties are configured by adding a bean of type `DbSchedulerOverrides` to your Spring context.
 
 ```java
 @Bean
@@ -464,13 +464,7 @@ DbSchedulerOverrides dbSchedulerOverrides() {
 }
 ```
 
-Anything left unset falls back to the corresponding property, and then to the library default. To set something programmatically that does have a property, or to reach a `SchedulerBuilder` option the starter does not surface, use the `customizeBuilder` escape hatch — it is applied last, so it wins over the properties:
-
-```java
-DbSchedulerOverrides.builder()
-    .customizeBuilder(builder -> builder.registerShutdownHook())
-    .build();
-```
+Anything left unset falls back to the corresponding property, and then to the library default. To set something programmatically that does have a property, or to reach a `SchedulerBuilder` option the starter does not surface, use the `customizeBuilder` escape hatch.
 
 **NB:** `DbSchedulerOverrides` replaces `DbSchedulerCustomizer` as of 17.0.0. An existing `DbSchedulerCustomizer` bean is still honoured, but it is deprecated and support for it will be removed in a future version.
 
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -452,7 +452,27 @@ For Spring Boot applications, there is a starter `db-scheduler-spring-boot-start
 
 ### Configuration options
 
-Configuration is mainly done via `application.properties`. Configuration of scheduler-name, serializer and executor-service is done by adding a bean of type `DbSchedulerCustomizer` to your Spring context.
+Configuration is mainly done via `application.properties`. Settings that cannot be expressed as properties — scheduler-name, serializer, executor-services, `JdbcCustomization` and `DataSource` — are configured by adding a bean of type `DbSchedulerOverrides` to your Spring context.
```
kagkarlsson: Remove listing all: `— scheduler-name, serializer, executor-services, `JdbcCustomization` and `DataSource` —` to avoid having to maintain the comment for new props

README.md
```diff
@@ -452,7 +452,27 @@ For Spring Boot applications, there is a starter `db-scheduler-spring-boot-start
...
+DbSchedulerOverrides dbSchedulerOverrides() {
+  return DbSchedulerOverrides.builder()
+      .schedulerName(new SchedulerName.Fixed("scheduler-1"))
+      .serializer(new JacksonSerializer())
+      .build();
+}
+```
+
+Anything left unset falls back to the corresponding property, and then to the library default. To set something programmatically that does have a property, or to reach a `SchedulerBuilder` option the starter does not surface, use the `customizeBuilder` escape hatch — it is applied last, so it wins over the properties:
+
```
kagkarlsson: Remove 

`— it is applied last, so it wins over the properties:

```java
DbSchedulerOverrides.builder()
    .customizeBuilder(builder -> builder.registerShutdownHook())
    .build();
```
`

db-scheduler-spring-boot-starter-parent/db-scheduler-spring-common/src/main/java/com/github/kagkarlsson/scheduler/boot/config/DbSchedulerCustomizer.java
```diff
@@ -24,7 +24,13 @@
 /**
  * Provides functionality for customizing various aspects of the db-scheduler configuration that is
  * not easily done with properties.
+ *
+ * @deprecated in favour of {@link DbSchedulerOverrides}, which covers the same settings as an
```
kagkarlsson: Drop `, which covers the same settings as an
 *     immutable value object rather than an interface to implement. Expose a {@code
 *     DbSchedulerOverrides} bean instead; a bean of this type is still honoured, but support will
 *     be removed in a future version.`

PR #914, opened 2026-08-29
fix: add Automatic-Module-Name to the Spring starter jars

kagkarlsson:
Set module-names so that spring boot starter can be consumed from a JPMS project. Currently, with no manifest entry, the module name is derived from the file name, and `db-scheduler-spring-boot-4-starter` derives to `db.scheduler.spring.boot.4.starter` — `4` is not a legal Java identifier, so `jar --describe-module` fails outright.

Fixes #909
Fixes #498

---
This PR was prepared with Claude Code.

PR #917, opened 2026-08-29
fix: swap module names between the Boot 3 and Boot 4 starters

kagkarlsson:
Follow-up to #914.

- Revert the module-name change for the Boot 3 starter — back to `db.scheduler.spring.boot.starter`.
- Rename the Boot 4 starter module to `com.github.kagkarlsson.scheduler.boot`.

---
This PR was prepared with Claude Code.

### DIFF
--- a/db-scheduler-spring-boot-starter-parent/db-scheduler-spring-boot3-starter/pom.xml
+++ b/db-scheduler-spring-boot-starter-parent/db-scheduler-spring-boot3-starter/pom.xml
@@ -157,8 +157,6 @@
     </pluginManagement>
 
     <plugins>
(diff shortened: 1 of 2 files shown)


db-scheduler-spring-boot-starter-parent/db-scheduler-spring-boot3-starter/pom.xml
```diff
@@ -157,12 +157,14 @@
     </pluginManagement>
 
     <plugins>
+      <!-- Freezes the name this jar has derived from its filename since 6.0; the clean
+           com.github.kagkarlsson.scheduler.boot goes to the Boot 4 starter, which outlives this one -->
```
kagkarlsson: Remove

